PR #54, opened 2019-06-17
Add Android API docs

threeve:


docs/package.json
```diff
@@ -5,6 +5,7 @@
   "private": true,
   "scripts": {
     "build": "eleventy",
+    "deploy": "gh-pages -d dist",
```
wes566: does this mean we'll start using a `gh-pages` branch for the docs?

threeve: yeah, that's the plan. Do you think we should do something else? @wes566

wes566: nope, that totally works 👍

Issue #51, opened 2019-06-07
Callback does not support Encodable parameter.


sfdctaka:
iOS callback does not currently support `Encodable` as a parameter.  `Encodable` is only supported now as return value.  

`Encodable` should be allowed in the parameter too.

PR #60, opened 2019-06-25
Support non-primitive types in callback parameters.

sfdctaka:
Non-primitive types can be passed to callback parameters. Following modifications were made:
* In Swift, `bind` methods constraint generic types to `Encodable` so that objects and primitives that can be JSON-ified be passed in.
* In Kotlin, annotation processor was updated to pass `JSONSerializable` object as is.
* Updated mocha tests.
* Updated demo-app.
* Updated binder tests in Swift.

Fixes https://github.com/salesforce/nimbus/issues/51.

### DIFF
--- a/platforms/android/nimbus/src/androidTest/java/com/salesforce/nimbus/MochaTests.kt
+++ b/platforms/android/nimbus/src/androidTest/java/com/salesforce/nimbus/MochaTests.kt
@@ -23,7 +23,7 @@ import java.util.concurrent.TimeUnit
 @RunWith(AndroidJUnit4::class)
 class MochaTests {
 
-    data class MochaMessage(val stringField: String = "This is a string", val intField: Int = 42) : JSONSerializable {
+    data class MochaMessage(var stringField: String = "This is a string", var intField: Int = 42) : JSONSerializable {
         override fun stringify(): String {
             val jsonObject = JSONObject()
             jsonObject.put("stringField", stringField)
@@ -73,9 +73,12 @@ class MochaTests {
         val webView = activityRule.activity.webView
         val testBridge = MochaTestBridge(webView)
 
+        val bridge = NimbusBridge("file:///android_asset/test-www/index.html")
+
         runOnUiThread {
             webView.addJavascriptInterface(testBridge, "mochaTestBridge")
-            webView.loadUrl("file:///android_asset/test-www/index.html")
+            bridge.add(CallbackTestExtensionBinder(CallbackTestExtension()))
+            bridge.attach(webView)
         }
 
         testBridge.readyLatch.await(5, TimeUnit.SECONDS)
--- a/platforms/android/nimbus/src/androidTest/java/com/salesforce/nimbus/MochaTests.kt
+++ b/platforms/android/nimbus/src/androidTest/java/com/salesforce/nimbus/MochaTests.kt
@@ -23,7 +23,7 @@ import java.util.concurrent.TimeUnit
 @RunWith(AndroidJUnit4::class)
 class MochaTests {
 
-    data class MochaMessage(var stringField: String = "This is a string", var intField: Int = 42) : JSONSerializable {
+    data class MochaMessage(val stringField: String = "This is a string", val intField: Int = 42) : JSONSerializable {
         override fun stringify(): String {
             val jsonObject = JSONObject()
             jsonObject.put("stringField", stringField)
--- a/platforms/apple/Sources/Nimbus/Callback.swift
+++ b/platforms/apple/Sources/Nimbus/Callback.swift
@@ -30,9 +30,9 @@ class Callback: Callable {
 
     func call(args: [Any]) throws -> Any {
         var jsonString: String = "[]"
-        if let bbb = args as? [EncodableValue] {
+        if let encodables = args as? [EncodableValue] {
             let jsonEncoder = JSONEncoder()
-            let jsonData = try? jsonEncoder.encode(bbb)
+            let jsonData = try? jsonEncoder.encode(encodables)
             jsonString = String(data: jsonData!, encoding: .utf8)!
         }
 
--- a/platforms/apple/Sources/Nimbus/Callback.swift
+++ b/platforms/apple/Sources/Nimbus/Callback.swift
@@ -30,17 +30,25 @@ class Callback: Callable {
 
     func call(args: [Any]) throws -> Any {
         var jsonString: String = "[]"
-        if let encodables = args as? [EncodableValue] {
+        if let encodables = args as? [Encodable] {
             let jsonEncoder = JSONEncoder()
-            let jsonData = try jsonEncoder.encode(encodables)
+            let encodableValues = encodables.map { encodable in
+                return EncodableValue.value(encodable)
+            }
+            let jsonData = try jsonEncoder.encode(encodableValues)
             jsonString = String(data: jsonData, encoding: .utf8)!
+        } else {
+            // Parameters passed to callback are implied that they
+            // conform to Encodable protocol.  If for some reason
+            // any elements don't throw parameter error.
+            throw ParameterError()
         }
-
+        
         DispatchQueue.main.async {
             self.webView?.evaluateJavaScript("""
-                var jsonArgs = \(jsonString);
-                jsonArgs.map(arg => arg.v);
-                nimbus.callCallback('\(self.callbackId)', jsonArgs);
+                var jsonArgs = JSON.parse('\(jsonString)');
+                mappedJsonArgs = jsonArgs.map(arg => arg.v);
+                nimbus.callCallback('\(self.callbackId)', mappedJsonArgs);
             """)
         }
         return ()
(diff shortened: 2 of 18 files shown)


platforms/apple/Sources/Nimbus/Callback.swift
```diff
@@ -29,11 +29,21 @@ class Callback: Callable {
     }
 
     func call(args: [Any]) throws -> Any {
-        let jsonData = try JSONSerialization.data(withJSONObject: args, options: [])
-        let jsonString = String(data: jsonData, encoding: .utf8)
+        var jsonString: String = "[]"
+        if let encodables = args as? [EncodableValue] {
+            let jsonEncoder = JSONEncoder()
+            let jsonData = try? jsonEncoder.encode(encodables)
+            jsonString = String(data: jsonData!, encoding: .utf8)!
```
threeve: This `try?` followed by force-unwrapping the result will result in a misleading error being thrown if encoding fails. It would probably be better to use `try` and let the encoding error propagate up here.

platforms/apple/Sources/Nimbus/Callback.swift
```diff
@@ -29,11 +29,21 @@ class Callback: Callable {
     }
 
     func call(args: [Any]) throws -> Any {
-        let jsonData = try JSONSerialization.data(withJSONObject: args, options: [])
-        let jsonString = String(data: jsonData, encoding: .utf8)
+        var jsonString: String = "[]"
+        if let encodables = args as? [EncodableValue] {
```
threeve: If the args are not encodable, what will happen here? It seems like the function will attempt to call through but will end up passing no arguments?

platforms/apple/Sources/Nimbus/Callback.swift
```diff
@@ -29,11 +29,21 @@ class Callback: Callable {
...
+            let jsonData = try? jsonEncoder.encode(encodables)
+            jsonString = String(data: jsonData!, encoding: .utf8)!
+        }
+
         DispatchQueue.main.async {
             self.webView?.evaluateJavaScript("""
-            nimbus.callCallback('\(self.callbackId)', \(jsonString!));
+                var args = [];
+                var jsonArgs = \(jsonString);
+                jsonArgs.forEach(arg => {
```
threeve: you should be able to use Array.map here rather than imperatively pushing values onto a new array:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

platforms/apple/Sources/NimbusTests/BinderTests.swift
```diff
@@ -67,13 +67,14 @@ class BinderTests: XCTestCase {
 
     func testBindUnaryWithUnaryCallback() {
         binder.bind(BindTarget.unaryWithUnaryCallback, as: "")
+        //binder.bind(BindTarget.unaryWithUnaryCallbackThrows, as: "")
```
threeve: remove?

platforms/android/nimbus/src/androidTest/java/com/salesforce/nimbus/MochaTests.kt
```diff
@@ -23,7 +23,7 @@ import java.util.concurrent.TimeUnit
 @RunWith(AndroidJUnit4::class)
 class MochaTests {
 
-    data class MochaMessage(val stringField: String = "This is a string", val intField: Int = 42) : JSONSerializable {
+    data class MochaMessage(var stringField: String = "This is a string", var intField: Int = 42) : JSONSerializable {
```
threeve: was this change necessary?

platforms/apple/Sources/Nimbus/Callback.swift
threeve: There seems to be an implicit requirement here that `Callback` now must be called with `[EncodableValue]`, and this has resulted in quite a bit of churn in other code. I wonder if we might be able to express this with less churn by changing the expectation to `Callback` accepting an `[Encodable]` and adding a case to `EncodableValue` for handling `[Encodable]`? We might even be able to lift that requirement all the way to `Callable` which would be nice because then it's part of the actual interface and becomes an explicit constraint rather than implicit.

platforms/apple/Sources/NimbusTests/BinderTests.swift
```diff
@@ -160,7 +163,7 @@ class BinderTests: XCTestCase {
         let callback: BindTarget.UnaryCallback = { value in
             expecter.fulfill()
         }
-        XCTAssertThrowsError(try binder.callable?.call(args: [42, make_callable(callback)]))
```
threeve: all these changes to remove `make_callable` confuse me... can you shed some light on why you did this throughout the test cases?

platforms/apple/Sources/NimbusTests/BinderTests.swift
```diff
@@ -429,8 +442,8 @@ enum BindError: Error {
 class BindTarget {
     private(set) var called = false
 
-    typealias UnaryCallback = (Int) -> Void
```
threeve: I'm also a bit confused by this change, can you walk me through why these callback signatures changed?

test-www/test/globals.d.ts
```diff
@@ -5,9 +5,20 @@
...
+}
+
 interface MochaTestBridge {
   ready(): void;
   sendMessage(name: string, includeParam: boolean): void;
 }
 
-declare var mochaTestBridge: MochaTestBridge;
\ No newline at end of file
+interface CallbackTestExtension {
```
threeve: the interfaces you added here could probably be placed in callback-encodable-tests.ts since they are specific to those test cases.

platforms/apple/Sources/Nimbus/Callback.swift
sfdctaka: In `Binder.swift` callbacks are called with arguments(`args`) passed to them wrapped in `EncodableValue`. So in theory all elements in `args` should be `EncodableValue` but that is not easily perceived as the method signature of `call` which it inherits and defines from protocol `Callable`. If this assumption is valid, then, we can add an additional check such as if elements of `args` match with yielding array of JSON objects(after encoding elements of `args` individually instead of the whole array in one shot as I do it now).

platforms/android/nimbus/src/androidTest/java/com/salesforce/nimbus/MochaTests.kt
sfdctaka: I had to do this because for callback that takes to `MochaMessage`s as parameters I wanted to change the second one's property values so that they're different from the first `MochaMessage`.

platforms/apple/Sources/Nimbus/Callback.swift
sfdctaka: Let me explore this.  Ping you tomorrow.

platforms/apple/Sources/NimbusTests/BinderTests.swift
sfdctaka: I removed `make_callable` because with it it produces `((Encodable)->()) throws -> Void` where as what I wanted was `(Encodable) -> ()`.

platforms/apple/Sources/NimbusTests/BinderTests.swift
sfdctaka: I changed this because in `Binder`'s `bind` method:
image
in line 61 I wrap `cb0` in `EncodableValue`.  

Then when the callback is called here and execution goes through line 51, then, type conversion to `A0` fails and was throwing `ParameterError`.
image

Same applies for `BinaryCallback` in `BindTarget`.

platforms/android/nimbus/src/androidTest/java/com/salesforce/nimbus/MochaTests.kt
sfdctaka: I got why you asked `val` vs `var`.  In the commit I had not included the file that requires this change.  It's coming in a follow-up commit.

platforms/apple/Sources/Nimbus/Callback.swift
```diff
@@ -29,11 +29,26 @@ class Callback: Callable {
     }
 
     func call(args: [Any]) throws -> Any {
-        let jsonData = try JSONSerialization.data(withJSONObject: args, options: [])
-        let jsonString = String(data: jsonData, encoding: .utf8)
+        var jsonString: String = "[]"
+        if let encodables = args as? [Encodable] {
+            let jsonEncoder = JSONEncoder()
+            let encodableValues = encodables.map { encodable in
+                return EncodableValue.value(encodable)
```
sfdctaka: @threeve I tried to extend `EncodableValue` to be top-level class for `[Encodable]`.  However, I couldn't get `encode(...)` method to work and decided to wrap the elements of `args` here instead. Will ping you again tomorrow and see if there's a better approach.

threeve: It's possible to write such an `encode` method, let's work through it together.

PR #61, opened 2019-07-02
Rewrite demo app with vanilla web components

threeve:


demo-www/src/components/device-info.ts
```diff
@@ -0,0 +1,36 @@
...
+    let shadowRoot = this.attachShadow({ mode: "open" });
+    shadowRoot.appendChild(template.content.cloneNode(true));
+  }
+
+  public connectedCallback(): void {
+    DeviceExtension.getDeviceInfo().then((info: DeviceInfo): void => {
+      console.log(JSON.stringify(info));
+      let shadowRoot = this.shadowRoot;
+      if (shadowRoot === null) return;
+      let slot = shadowRoot.querySelector("slot");
```
wes566: this totally works but a suggestion, you could just make a private field to track your slot and set that in the ctor, then you don't need to check if shadowroot is null or do query selection on each connected callback... I have an example [in a repo here](https://github.com/wes566/infinite-carousel-wc/blob/master/lib/infinite-carousel-wc.ts#L52).

demo-www/public/index.html
```diff
@@ -1,15 +1,21 @@
...
-    <title>This is an lwc app</title>
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <title>Nimbus Demo</title>
+    <script src="bundle.js"></script>
   </head>
   <body>
-    <p>Hello.</p>
-    <div id="app"></div>
-    <script src="/bundle.js"></script>
+    <nimbus-demo-app>
```
wes566: since `nimbus-demo-app` and `nimbus-header` are just slots without any other features/css then you might as well put `div`s here, I think the extra empty custom elements don't add anything to the demo (if anything they kind of introduce unnecessary complexity)... using web components doesn't mean you have to make everything a custom element when built-in elements will suffice 🤷‍♂

threeve: Yeah, perhaps planning a little too far ahead... my thought was to turn the `nimbus-demo-app` component into an accordion or master/detail to drill into the various subcomponents that demo the built-in extensions. I'll simplify for now though.

threeve:
These changes were included as part of #69

PR #85, opened 2019-09-11
Add support iOS Foundation type.  Updated Android tests.

sfdctaka:
This is a draft PR to get feedback.

Added support to allow iOS's Foundation type(specifically, `NSArray` and `NSDictionary`) as return type from native methods or as parameters in the callback functions.

In doing so it proliferated unit and mocha tests.  A better approach might have been to encapsulate `NSArray` and `NSDictionary` into an enum, like we do for `EncodableValue`, so a union type could be created.  This way we can reduce matching signatures and consequently the number of tests.  However, I wasn't quite sure if this was possible.  @threeve , we spoke briefly on this bug/enhancement and you were saying you might have an idea on how to approach.  Were you also thinking wrapping the two Foundation types in enum or a class? Please advise and I'll happy to update this PR to reduce the lines of code that were added.

### DIFF
--- a/platforms/android/nimbus/src/androidTest/java/com/salesforce/nimbus/CallbackTestExtension.kt
+++ b/platforms/android/nimbus/src/androidTest/java/com/salesforce/nimbus/CallbackTestExtension.kt
@@ -1,9 +1,7 @@
 package com.salesforce.nimbus
 
-
 import org.json.JSONArray
 import org.json.JSONObject
-
 
 @Extension(name = "callbackTestExtension")
 class CallbackTestExtension : NimbusExtension {
@@ -34,11 +32,10 @@ class CallbackTestExtension : NimbusExtension {
     }
 
     @ExtensionMethod
-    fun callbackWithPrimitiveAndArrayParams(arg:(param0: Int, param1: JSONArray) -> Unit) {
+    fun callbackWithPrimitiveAndArrayParams(arg: (param0: Int, param1: JSONArray) -> Unit) {
         var ja = JSONArray(listOf("one", "two", "three"))
         arg(777, ja)
     }
-
 
     @ExtensionMethod
     fun callbackWithPrimitiveAndDictionaryParams(arg: (param0: Int, param1: JSONObject) -> Unit) {
--- a/platforms/android/nimbus/src/androidTest/java/com/salesforce/nimbus/CallbackTestExtension.kt
+++ b/platforms/android/nimbus/src/androidTest/java/com/salesforce/nimbus/CallbackTestExtension.kt
@@ -1,5 +1,10 @@
 package com.salesforce.nimbus
 
+
+import org.json.JSONArray
+import org.json.JSONObject
+
+
 @Extension(name = "callbackTestExtension")
 class CallbackTestExtension : NimbusExtension {
     @ExtensionMethod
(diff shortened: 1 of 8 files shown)


platforms/android/nimbus/src/androidTest/java/com/salesforce/nimbus/CallbackTestExtension.kt
```diff
@@ -27,4 +32,75 @@ class CallbackTestExtension : NimbusExtension {
     fun callbackWithPrimitiveAndUddtParams(arg: (param0: Int, param1: MochaTests.MochaMessage) -> Unit) {
         arg(777, MochaTests.MochaMessage())
     }
+
+    @ExtensionMethod
+    fun callbackWithPrimitiveAndArrayParams(arg:(param0: Int, param1: JSONArray) -> Unit) {
```
sfdctaka: @threeve Something interesting I found along the way is when I was updating Kotlin code and declaring callback parameters as `List<T>` or `Array<T>` or `HashMap<String, Int>` these would not be converted to Javascript array and dictionary(object).  They started working once I started using `JSONArray` and `JSONObject`.  I'm wondering if we need to do a similar exercise that I did here for Swift in Kotlin to support the `List<T>`, `Array<T>`, and `HashMap<String, Int>` .

threeve: Android supports ArrayList and HashMap on the input param side, just not as callback params yet. (see https://github.com/salesforce/nimbus/blob/master/platforms/android/nimbus-compiler/src/main/java/com/salesforce/nimbus/NimbusProcessor.kt#L211-L232)

If we could extend the built-in types to support our `JSONSerialization` interface that's probably the ideal, but the fallback would be to simply handle them explicitly when generating the callback binder code similar to how we do for params.

platforms/apple/Sources/Nimbus/Binder.swift
```diff
@@ -254,11 +512,559 @@ extension Binder {
...
+            }
+        }
+        let callable = make_callable(wrappedFunction)
+        bind(callable, as: name)
+    }
+
+    /**
+     Bind the specified function to this connection.
+     */
+    public func bind<A0, A1, A2, A3, CBD0: NSDictionary, CBD1: NSDictionary>(_ function: @escaping (Target) -> (A0, A1, A2, A3, @escaping (CBD0, CBD1) -> Void) throws -> Void, as name: String) {
```
threeve: This explosion of `bind` overloads definitely feels like it crosses the threshold to where this file is not maintainable by hand any more.

We have a few potential paths forward that we can discuss:

- Figure out how to rewrite the bind overloads to reduce their numbers. One way I've been thinking of doing this would be to take every input function, curry it, then apply argument coercion to each curried arg. This might be able to take us from the N x M situation we have now to an N + M. I'm not sure this will work with callbacks though due to how swift handles callback types in generics

- Autogenerate the binds using a template and code generation tool like sourcery. This would at least reduce the code that needs to be manually maintained to an acceptable level, and we'd just have it autogenerate the menagerie of bind overloads that need to exist

- Create a binder generator more like android. This would be a tool that iterates over the swift code and generates a bespoke binder for each extension, similar to how android handles @Extension and @ExtensionMethod annotations

platforms/apple/Sources/Nimbus/Callback.swift
```diff
@@ -29,22 +29,36 @@ class Callback: Callable {
...
-            throw ParameterError()
+        let jsonEncoder = JSONEncoder()
+        let jsonArgs = try args.map { arg -> String in
+            if let encodable = arg as? Encodable {
+                let jsonData = try jsonEncoder.encode(EncodableValue.value(encodable))
+                let jsonString = String(data: jsonData, encoding: .utf8)!
+                return jsonString
+            } else if arg is NSArray || arg is NSDictionary {
+                // swiftlint:disable:next force_try
+                let data = try! JSONSerialization.data(withJSONObject: arg, options: [])
```
threeve: Just use `try` here, without the forcing. This method is `throws` so it's okay to let the error bubble up out of this method. Ideally we want to not add any additional force trys and get rid of the few existing cases we do have.

PR #88, opened 2019-10-01
Use iOS's Foundation types via sourcery

sfdctaka:
This is a follow-up PR to https://github.com/salesforce/nimbus/pull/87.  It brings #87 to equivalent of close PR #85 using Sourcery code generation. The diff between this PR and #87 should suffice to show the slight improvement there in writing repeated, almost similar bind code in iOS and its respective tests.

I did not mix this code generation improvement using Sourcery with currying improvement as I couldn't put together a solution that both improvements would work together.

### DIFF
--- a/platforms/android/nimbus/src/androidTest/java/com/salesforce/nimbus/CallbackTestExtension.kt
+++ b/platforms/android/nimbus/src/androidTest/java/com/salesforce/nimbus/CallbackTestExtension.kt
@@ -1,5 +1,8 @@
 package com.salesforce.nimbus
 
+import org.json.JSONArray
+import org.json.JSONObject
+
 @Extension(name = "callbackTestExtension")
 class CallbackTestExtension : NimbusExtension {
     @ExtensionMethod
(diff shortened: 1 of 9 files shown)


platforms/android/nimbus/src/androidTest/java/com/salesforce/nimbus/CallbackTestExtension.kt
```diff
@@ -27,4 +30,74 @@ class CallbackTestExtension : NimbusExtension {
     fun callbackWithPrimitiveAndUddtParams(arg: (param0: Int, param1: MochaTests.MochaMessage) -> Unit) {
         arg(777, MochaTests.MochaMessage())
     }
+
+    @ExtensionMethod
+    fun callbackWithPrimitiveAndArrayParams(arg: (param0: Int, param1: JSONArray) -> Unit) {
+        var ja = JSONArray(listOf("one", "two", "three"))
+        arg(777, ja)
+    }
+
```
sfdctaka: Adding the built-in types to support JSONSerialization interface as it was mentioned in #85 is left to do for another story.

PR #89, opened 2019-10-02
Change how nimbus is exported to avoid global nesting in iife module

threeve:


sfdctaka:
I'm just curious. Single quotes were replaced by double quotes. What was the motivation?

wes566:
> I'm just curious. Single quotes were replaced by double quotes. What was the motivation?

I'm guessing prettier missed some files earlier

threeve:
> I'm just curious. Single quotes were replaced by double quotes. What was the motivation?

Yes, it's just the result of prettier formatting the file when I edited it. At some point I will make sure prettier/lint is run on checkin and all files will get formatted according to the default style then but I didn't want to include all of that in this PR.

PR #93, opened 2019-10-08
Check bounds of args array prior to attempting conversion

threeve:


sfdctaka:
@threeve codecov picked up a file (callback.kt) that was not part of this PR and complaining about it.  Looking at now why it did that.

PR #95, opened 2019-10-09
Support passing freestanding functions to `bind`

### LINKED ISSUE #83: bind should accept functions that are already bound to an instance
Currently calling `bind` on a `Connection` expects an unbound, curried function that takes the instance as the first param:

```swift
class MyType: NimbusExtension {
    func method(arg1: String, arg2: Int) -> String {
    }
    public func bindToWebView(webView: WKWebView) {
        let connection = webView.addConnection(to: self, as: "MyType")
        connection.bind(MyType.method, as: "method")
    }
}
```

However swift will vend bound functions from the current instance:
```swift
    public func bindToWebView(webView: WKWebView) {
        let connection = webView.addConnection(to: self, as: "MyType")
        let f = method // type of f => (String, Int) -> String
        connection.bind(f, as: "method")
    }
```
but this method currently fails to bind as it cannot find a suitable overload to call with the pre-bound method.

We should support both methods to avoid having to rebind functions to the instance when attaching to the webview.

threeve:
fixes #83

### DIFF
--- a/platforms/apple/Sources/Nimbus/Generated/Binders.generated.swift
+++ b/platforms/apple/Sources/Nimbus/Generated/Binders.generated.swift
@@ -16,8 +16,14 @@ extension Binder {
      Bind the specified function to this connection.
      */
     public func bind(_ function: @escaping (Target) -> () throws -> Void, as name: String) {
-        let boundFunction = function(target)
-        let callable = make_callable(boundFunction)
+        bind(function(target), as: name)
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


threeve:
Sadly I don't think this is completely possible to implement at the moment. When attempting to add test cases for functions with callbacks I hit an issue with escaping closures and generic overload ambiguity that doesn't appear to be resolvable. I will be closing this for now and may revisit at some point in the future.

PR #105, opened 2019-12-04
Remove cocoapods spec prefetch step

threeve:


threeve:
This is handled in @paultiarks latest PR

PR #106, opened 2019-12-11
Supporting promises by supplying closure

sfdctaka:
[RFC] https://github.com/salesforce/nimbus/issues/107

With this change Nimbus will support an additional way to translate native method to JS method that expects a promise a return type.  Existing way was to create a native method to have a return type(non-void).  This change allows a method to have no return type.  Instead, it can be explicitly set as a parameter(iOS) or in annotation(Android) that a trailing closure should be used to resolve or reject promise in JS.

In iOS, `bind` is where a trailing closure can be explicitly set to be used for a callback or for a promise.

In Android, kotlin annotation is used.  Parameters set in this annotation is what's considered as metadata to nimbus extension's method's "metadata". A metadata can be a directive, like `trailingClosure = TrailingClosure.PROMISE`, which basically indicates "if there's a trailing closure then create a JS function with promise". To pass metadata to JS side for bootstrap a method was created, `getExtensionMetadata()`.  Currently it only has one metadata, `trailingClosuresToPromisify`, but it can be used in the future to pass any metadata to the JS side.

### DIFF
--- a/packages/nimbus-bridge/src/index.ts
+++ b/packages/nimbus-bridge/src/index.ts
@@ -61,7 +61,6 @@ class Nimbus {
     let dest: any = {};
     let trailingClosuresAsPromises: Array<string> = [];
     if (src['getExtensionMetadata'] !== undefined) {
-      console.log('ccc');
       let extensionMetadata = src.getExtensionMetadata.call(src);
       if (extensionMetadata !== undefined) {
         extensionMetadata = JSON.parse(extensionMetadata)
--- a/packages/nimbus-bridge/src/index.ts
+++ b/packages/nimbus-bridge/src/index.ts
@@ -59,22 +59,43 @@ class Nimbus {
 
   public promisify = (src: any) => {
     let dest: any = {};
+
+    let methodsWithClosuresToPromisify = "";
+    if (src['getMethodsWithClosuresToPromisify'] !== undefined) {
+      methodsWithClosuresToPromisify = src.getMethodsWithClosuresToPromisify.call(src)
+    }
+
+    let methodsWithClosuresToPromisifyArray = methodsWithClosuresToPromisify.split(',');
     Object.keys(src).forEach(key => {
       let func = src[key];
-      dest[key] = (...args: any[]) => {
-        args = this.cloneArguments(args);
-        args = args.map(arg => {
-          if (typeof arg === "object") {
-            return JSON.stringify(arg);
+      if (methodsWithClosuresToPromisifyArray.includes(key)) {
+        dest[key] = (...args: any[]) => {
+          let functionArgs = nimbus.cloneArguments(args);
+          return new Promise(function (resolve, reject) {
+            var promiseId = nimbus.uuidv4();
+            nimbus.promises[promiseId] = { resolve, reject };
+
+            func.call(src, ...functionArgs, promiseId);
+          });
+        };
+      }
+      else {
+        dest[key] = (...args: any[]) => {
+          args = this.cloneArguments(args);
+          args = args.map(arg => {
+            if (typeof arg === "object") {
+              return JSON.stringify(arg);
+            }
+            return arg;
+          });
+
+          let result = func.call(src, ...args);
+          if (result !== undefined) {
+            result = JSON.parse(result);
           }
-          return arg;
-        });
-        let result = func.call(src, ...args);
-        if (result !== undefined) {
-          result = JSON.parse(result);
-        }
-        return Promise.resolve(result);
-      };
+          return Promise.resolve(result);
+        };
+      }
     });
     return dest;
   };
--- a/packages/nimbus-bridge/src/index.ts
+++ b/packages/nimbus-bridge/src/index.ts
@@ -59,16 +59,21 @@ class Nimbus {
 
   public promisify = (src: any) => {
     let dest: any = {};
-
-    let methodsWithClosuresToPromisify = "";
-    if (src['getMethodsWithClosuresToPromisify'] !== undefined) {
-      methodsWithClosuresToPromisify = src.getMethodsWithClosuresToPromisify.call(src)
+    let trailingClosuresAsPromises: Array<string> = [];
+    if (src['getExtensionMetadata'] !== undefined) {
+      console.log('ccc');
+      let extensionMetadata = src.getExtensionMetadata.call(src);
+      if (extensionMetadata !== undefined) {
+        extensionMetadata = JSON.parse(extensionMetadata)
+        if (extensionMetadata['trailingClosuresAsPromises'] !== undefined) {
+          trailingClosuresAsPromises = extensionMetadata['trailingClosuresAsPromises'];
+        }
+      }
     }
 
-    let methodsWithClosuresToPromisifyArray = methodsWithClosuresToPromisify.split(',');
     Object.keys(src).forEach(key => {
       let func = src[key];
-      if (methodsWithClosuresToPromisifyArray.includes(key)) {
+      if (trailingClosuresAsPromises.includes(key)) {
         dest[key] = (...args: any[]) => {
           let functionArgs = nimbus.cloneArguments(args);
           return new Promise(function (resolve, reject) {
@@ -78,8 +83,7 @@ class Nimbus {
             func.call(src, ...functionArgs, promiseId);
           });
         };
-      }
-      else {
+      } else {
         dest[key] = (...args: any[]) => {
           args = this.cloneArguments(args);
           args = args.map(arg => {
--- a/platforms/android/nimbus-compiler/build.gradle
+++ b/platforms/android/nimbus-compiler/build.gradle
@@ -4,6 +4,7 @@ apply plugin: 'kotlin'
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
+    implementation 'org.json:json:20180813'
     implementation project(':nimbus-annotations')
     implementation 'com.squareup:javapoet:1.11.1'
 }
--- a/platforms/android/nimbus-compiler/src/main/java/com/salesforce/nimbus/NimbusProcessor.kt
+++ b/platforms/android/nimbus-compiler/src/main/java/com/salesforce/nimbus/NimbusProcessor.kt
@@ -111,6 +111,8 @@ class NimbusProcessor : AbstractProcessor() {
                         .addStatement("this.\$N = \$N", "webView", "webView")
                         .build())
 
+                var closuresToPromisify = mutableListOf<String>()
+
                 methods.forEach {
                     val methodElement = it as ExecutableElement
 
@@ -123,6 +125,11 @@ class NimbusProcessor : AbstractProcessor() {
 
                     val arguments = mutableListOf<String>()
                     var argIndex = 0
+
+                    val promisifyClosure = methodElement.getAnnotation(ExtensionMethod::class.java).promisifyClosure
+                    if (promisifyClosure) {
+                        closuresToPromisify.add(methodElement.simpleName.toString())
+                    }
 
                     methodElement.parameters.forEach {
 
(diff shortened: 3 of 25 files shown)


platforms/android/nimbus-compiler/src/main/java/com/salesforce/nimbus/NimbusProcessor.kt
```diff
@@ -179,17 +186,20 @@ class NimbusProcessor : AbstractProcessor() {
...
-                                                    CodeBlock.builder()
-                                                            .add("if (webView != null) {\n")
-                                                            .indent()
-                                                            .addStatement("callJavascript(\$N, \$S, \$N, null)", "webView", "nimbus.callCallback2", "args")
-                                                            .unindent()
-                                                            .add("}\n")
-                                                            .addStatement("return null")
-                                                            .build()
-                                            )
+                                    var javascriptMethodToCall = "nimbus.callCallback2"
```
jamie-houston: nit: You can make this immutable
```
val javascriptMethodToCall = if (promisifyClosure) "nimbus.resolvePromise" else "nimbus.callCallback2"
```

platforms/android/nimbus-compiler/src/main/java/com/salesforce/nimbus/NimbusProcessor.kt
```diff
@@ -272,10 +273,41 @@ class NimbusProcessor : AbstractProcessor() {
...
+                                throw Exception("Method with closure to be promisified can not have a return type")
+                            } else {
+                                if (it.returnType.toString().equals("java.lang.String")) {
+                                    methodSpec.addStatement("return \$T.quote(target.\$N($argsString))",
+                                        ClassName.get("org.json", "JSONObject"),
+                                        it.simpleName.toString())
+                                } else {
+
+                                    val supertypes = processingEnv.typeUtils.directSupertypes(it.returnType)
+                                    var found = false
```
jamie-houston: Maybe out of scope for this PR, but there's a lot of redundency in this class.  Also, we could simplify and make found immutable by
```
val found = supertypes.any{ superType -> superType.toString().equals("com.salesforce.nimbus.JSONSerializable")}
```

platforms/android/nimbus-compiler/src/main/java/com/salesforce/nimbus/NimbusProcessor.kt
sfdctaka: ahhhh, kotlin can return conditionals results!

platforms/android/nimbus-compiler/src/main/java/com/salesforce/nimbus/NimbusProcessor.kt
sfdctaka: @jamie-houston  This was a companion draft PR for the RFC.  I'll give another round of revision here and try to remove the redundancy before making this PR an official one.

platforms/android/nimbus-compiler/src/main/java/com/salesforce/nimbus/NimbusProcessor.kt
```diff
@@ -288,6 +320,23 @@ class NimbusProcessor : AbstractProcessor() {
...
+                                    AnnotationSpec.builder(ClassName.get("android.webkit", "JavascriptInterface"))
+                                            .build())
+                            .addModifiers(Modifier.PUBLIC)
+                            .returns(TypeName.get(String::class.java))
+
+                    methodSpec.addStatement("java.util.List<String> closureNames = new java.util.ArrayList<>()")
+                    closuresToPromisify.forEach { it ->
+                        methodSpec.addStatement("closureNames.add(\"\$N\")", it)
+                    }
+                    val commaSeparatedClosures = closuresToPromisify.joinToString(separator = ",");
```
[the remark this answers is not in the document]
sfdctaka: @threeve I tried passing back `String[]` and `JSONArray` but instead went for return an object(dict) with an array.  This object that gets returned by `getExtensionMetadata()` can be used to pass any metadata like a property bag.  There, there is a property `trailingClosuresToPromisify`.  Also, I went for this because `@JavascriptInterface` couldn't translate `String[]` as JSON array.

sfdctaka:
@paultiarks @threeve I updated the PR and would appreciate if you can check it out again. Thanks!

platforms/android/nimbus/src/main/java/com/salesforce/nimbus/extensions/DeviceExtension.kt
```diff
@@ -46,4 +47,9 @@ class DeviceExtension(context: Context) : NimbusExtension {
     fun getDeviceInfo(): DeviceInfo {
         return cachedDeviceInfo
     }
+
+    @ExtensionMethod(trailingClosure = TrailingClosure.PROMISE)
+    fun getDeviceDesignLocation(arg: (param0: String) -> Unit) {
+        arg("Mountain View, CA")
```
threeve: Perhaps we shouldn't include this method on DeviceExtension just to test trailing closure promises. There are lots of different Android device manufacturers so this value isn't really correct for most of them.

platforms/apple/Sources/Nimbus/Connection.swift
```diff
@@ -80,15 +85,19 @@ public class Connection<C>: Binder {
...
             do {
                 // walk args, converting callbacks into Callables
-                let args = args.map { arg -> Any in
+                var args = args.map { arg -> Any in
                     switch arg {
                     case let dict as NSDictionary:
                         if let callbackId = dict["callbackId"] as? String {
-                            return Callback(webView: webView!, callbackId: callbackId)
+                            if promisifyCallback {
+                                fatalError("Trailing closure that is promsified can not have callback ID")
```
threeve: This should throw or something, we don't want to `fatalError` in library code.

platforms/android/nimbus-compiler/src/main/java/com/salesforce/nimbus/NimbusProcessor.kt
```diff
@@ -365,4 +336,30 @@ class NimbusProcessor : AbstractProcessor() {
             element
         )
     }
+
+    private fun addReturnStatementToMethod(element: ExecutableElement, methodSpec: MethodSpec.Builder, argsString: String, isPromise: Boolean): Unit {
+        val supertypes = processingEnv.typeUtils.directSupertypes(element.returnType)
+        var found = false
+        for (supertype in supertypes) {
```
jamie-houston: this can be simplified to
``` val found = supertypes.any{ supertype -> supertype.toStrig().equals("com.salesforce.nimbus.JSONSerializable") }
```

So `found` is immutable and loop exits if one type is found.

platforms/android/nimbus/src/main/java/com/salesforce/nimbus/extensions/DeviceExtension.kt
```diff
@@ -2,10 +2,7 @@ package com.salesforce.nimbus.extensions
 
 import android.content.Context
 import android.content.pm.PackageManager
-import com.salesforce.nimbus.Extension
-import com.salesforce.nimbus.ExtensionMethod
-import com.salesforce.nimbus.JSONSerializable
-import com.salesforce.nimbus.NimbusExtension
+import com.salesforce.nimbus.*
```
jamie-houston: nit: Don't use wildcard imports.  Not sure how others feel about it though.

platforms/android/nimbus-compiler/build.gradle
```diff
@@ -4,6 +4,7 @@ apply plugin: 'kotlin'
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
+    implementation 'org.json:json:20180813'
```
jamie-houston: probably doesn't matter but this is an old version.  LDS-Android is using 20090211

platforms/android/nimbus-compiler/src/main/java/com/salesforce/nimbus/NimbusProcessor.kt
```diff
@@ -316,4 +336,30 @@ class NimbusProcessor : AbstractProcessor() {
...
+            if (supertype.toString().equals("com.salesforce.nimbus.JSONSerializable")) {
+                found = true
+            }
+        }
+
+        if (found) {
+            methodSpec.returns(String::class.java)
```
jamie-houston: nit: this conditional could set the statement, and then after
```
methodSpec.addStatement(statement, element.simpleName.toString())
```
not sure if the toString is necessary on it either.

platforms/apple/Sources/Nimbus/Connection.swift
```diff
@@ -43,11 +43,12 @@ public class Connection<C>: Binder {
                 let params = message.body as? NSDictionary,
                 let method = params["method"] as? String,
                 let args = params["args"] as? [Any],
-                let promiseId = params["promiseId"] as? String else {
+                let promiseId = params["promiseId"] as? String,
+                let promisifyCallback = params["promisifyCallback"] as? Bool else {
```
jamie-houston: nit: format/alignment of this ` else { return }` is confusing

platforms/apple/Sources/Nimbus/Connection.swift
```diff
@@ -56,20 +57,24 @@ public class Connection<C>: Binder {
     /**
      Bind the callable object to a function `name` under this conenctions namespace.
      */
-    public func bind(_ callable: Callable, as name: String) {
+    public func bind(_ callable: Callable, as name: String, trailingClosure: TrailingClosure) {
         bindings[name] = callable
+        var forPromisifiedClosure = "false"
```
jamie-houston: can this be simplified to something like
```
let forPromisifiedClosure = TrailingClosure.promise == trailingClosure
```
Not sure how to do that in swift.

platforms/android/nimbus/src/main/java/com/salesforce/nimbus/extensions/DeviceExtension.kt
sfdctaka: @threeve  I agree, Mountain View is a bit inaccurate to be the sole location of device manufacturer.  I am ready to remove this.  The intent of putting this code here was more for a quick usage example of trailing closure promises.  I think the unit tests should that well enough.

platforms/android/nimbus-compiler/build.gradle
sfdctaka: @jamie-houston It's 2018, not sure if you wanted me to use 2009? Here are versions I see in Maven https://mvnrepository.com/artifact/org.json/json

platforms/apple/Sources/Nimbus/Connection.swift
sfdctaka: @jamie-houston I'll put the `return` in the same line as the last `let`.

platforms/apple/Sources/Nimbus/Connection.swift
sfdctaka: `TrailingClosure.promise` is an enum and this is the only style we can use in Swift for now or otherwise I would have to write exhaustive `switch` statements.  It's something I personally feel wrong too in case but the current way is the smallest way to do the comparison.

platforms/android/nimbus-compiler/build.gradle
jamie-houston: haha... yikes.  Look like I need to use your version.  Weird... I use a tool to get latest versions of dependencies.  Looks like I found a bug.

jamie-houston: BTW, latest version is 20190722 but if I'm using a 10 year old version.... I'll just slowly make my exit....

packages/nimbus-bridge/src/index.ts
```diff
@@ -59,22 +59,47 @@ class Nimbus {
 
   public promisify = (src: any) => {
     let dest: any = {};
+    let trailingClosuresAsPromises: Array<string> = [];
+    if (src['getExtensionMetadata'] !== undefined) {
+      console.log('ccc');
```
threeve: I'd remove this console.log

sfdctaka: 😓

PR #109, opened 2019-12-17
Subprojects weren't getting lint and publishing gradles applied properly on afterEvaluate.

sfdctaka:


threeve:
Can you provide some more info on what the failure mode was prior to this change? IIRC we moved these all into the root `build.gradle` to work around issues when consuming the android targets via submodule. If we move these plugins back into the individual project `build.gradle` we have probably reintroduced that issue and made it so consumption via submodule is harder or not possible.

sfdctaka:
@threeve The error was `Repository name, package name or version name are null for project nimbus-compiler` when I `./gradlew bintrayUpload` at Android root.  Other libraries would upload but not `nimbus-compiler`. I checked with @jamie-houston and he told me that LDS Android had the same issue and moved the lint and publishing gradles to per project.  

I don't mind closing this PR and keep the way it was.  However, that won't guarantee consumption via submodule is still working.  Anyway, do you think doing one offs locally, as in this PR, to upload to bintray suffices for now?

threeve:
I definitely wouldn't want to have to make this change locally every time we do a release, but I want to think through whether there are any other implications is all. We may just need to make the statement that integration via submodule is not supported for Android and direct people to the bintray packages. @jamie-houston @jhorst11 is this a potential issue with those composite builds that you experimented with or should this work ok there?

PR #112, opened 2020-01-03
Compile nimbus runtime module to es5 to support old Android versions

threeve:


threeve:
@wes566 I agree with you, I changed it up so the iife module uses es5 and the es modules uses es6. Does this seem better?

wes566:
@threeve This looks good.  I forgot that we already use rollup and output an iife and an es module... guess I should double check the codebase before reviewing and leaving comments 😆

PR #115, opened 2020-01-08
Bump version

sfdctaka:


threeve:
Yeah, as Paul points out we should start keeping the versions in sync, which means just updating the Nimbus.podspec in the repo root as well.

Issue #116, opened 2020-01-13
[RFC] Supporting reject state for promise created with trailing closure


sfdctaka:
Nimbus currently has a way to turn a native method into Javascript method that returns a promise.  For example let's say that we have a method `foo` in iOS:
```
func foo(@escaping callback: (success: String)->Void) {...}
```
and a counterpart in Android:
```
func foo(callback: (success: String)->Unit) {...}
```
For these two methods to be converted into a method that returns a promise in Javascript we need to bind `foo`, for iOS, and indicate that the closure that was passed as a parameter is for promise:
```
connection.bind(foo, as: "foo", trailingClosure: .promise)
```
And for Android we need to annotate `foo` with:
```
@ExtensionMethod(trailingClosure = TrailingClosure.PROMISE)
func foo(callback:(success:String)-> Unit) {...}
```
In Javascript the above would translate into code like this:
```
const promise = foo()
```
But because this promise currently only supports the resolve state but not the reject state it is only then-able:
```
// Doable:
foo().then((success) -> {console.log(success);})
// Not doable:
foo().catch((error) -> {console.log(error);})
// Also not doable:
foo().then((success) -> {console.log(success);}).catch((error) -> {console.log(error);})
```

This RFC lists and explores 3 distinct approaches to support reject state in Nimbus.

**(1) Method that Throws**
If a method is declared to throw and is designated as a trailing closure,
in iOS:
```
func foo(@escaping callback: (success: String) -> ()) throws {...}
connection.bind(foo, as: "foo", trailingClosure: .promise)
```
in Android:
```
@Throws(IOException::class)
@ExtensionMethod(trailingClosure = TrailingClosure.PROMISE)
fun foo(callback: (success: String) -> Unit) {...}
```
then we can consider that promise for these methods will need to support both resolve and reject. All exceptions will be converted into 'something'(string message??).

Pro: Fairly easy to implement
Con: Exception thrown will lose information when it gets to Javascript as a string message. What type of data should reject state pass back??

**(2) Closure with Two Arguments**
If a method is passed a closure with two arguments and this closure is indicated as promise
in iOS:
```
func foo(@escaping callback: (success: String?, error: String?) -> ()) {...}
connection.bind(foo, as: "foo", trailingClosure: .promise)
```
in Android:
```
@ExtensionMethod(trailingClosure = TrailingClosure.PROMISE)
fun foo(callback: (success: String?, error: String?) -> ()) {...}
```
[...]

PR #118, opened 2020-01-15
RFC for supporting reject state in promise created with trailing closure.

sfdctaka:
This is a copy of https://github.com/salesforce/nimbus/issues/116 so we can keep a copy of RFC as a part of documentation and trace our decision.

### DIFF
--- a/docs/site/rfcs/supportRejectInPromiseByTrailingClosure.md
+++ b/docs/site/rfcs/supportRejectInPromiseByTrailingClosure.md
@@ -1,26 +1,28 @@
+# Supporting reject state for promise created with trailing closure
+
 Nimbus currently has a way to turn a native method into Javascript method that returns a promise.  For example let's say that we have a method `foo` in iOS:
-```
-func foo(@escaping callback: (success: String)->Void) {...}
+```swift
+func foo(@escaping callback: (success: String) -> Void) {...}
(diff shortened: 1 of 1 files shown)


sfdctaka:
Thanks for the review guys! I explored option 1 and I thought I had it but there was a problem when implementing it in Android.  Kotlin Annotation Processor in Nimbus(nimbus-compiler) is not able to see the annotation `@Throws(...)` when there is a closure passed as an argument to a method using `methodElement.thrownTypes`. This works fine for non-closure parameters. Unless nimbus-compiler can check the existence of `@Throws(...)` on a method there is no way to determine if a method throws exception.  So now I'm going to see how fast I can do option 2.

threeve:
my understanding is that Kotlin does not enforce checked exceptions, so in essense _any_ method could potentially throw. Do we need to require that it is annotated as `@Throws` or can we assume any callback _might_ throw and treat any caught exception as a promise rejection when using promises?

sfdctaka:
@threeve Ah, thanks for sharing this info with me! I was hung on need `@Throws`but let me try implementing it even without `@Throws` and assume that if the method is annotated with`@ExtensionMethod(trailingClosure = TrailingClosure.PROMISE)` it supports both resolve and reject.

PR #123, opened 2020-01-23
Android: preserve and extend JS objects

### LINKED ISSUE #122: Android should preserve and extend JS objects
To match the behavior on iOS, Android should cause existing JS objects to
be preserved.

E.g. if the sample `DeviceExtension` is added to the NimbusBridge, the
following code should work, as it does on iOS
```js
(function() {

const DeviceExtension = window.DeviceExtension || (window.DeviceExtension = {})

DeviceExtension.doSomethingElse = () => {
  // Once Nimbus is initialized, window.DeviceExtension
  // should have { getDeviceInfo, doSomethingElse }
  DeviceExtension.getDeviceInfo().then(info => {
    console.log(info)
  })
}

})();
```

Currently, the behavior depends on whether the above JS code
executes before or after Nimbus initialization.

For parity with iOS, both `getDeviceInfo` and `doSomethingElse` should
always be available after initialization, regardless of initialization order.

AdamLiechty:
Fixes #122 

To match the behavior on iOS, this change causes existing JS objects to
be preserved.

E.g. if the sample `DeviceExtension` is added to the NimbusBridge, the
following code should work, as it does on iOS
```js
(function() {

const DeviceExtension = window.DeviceExtension || (window.DeviceExtension = {})

DeviceExtension.doSomethingElse = () => {
  // Once Nimbus is initialized, window.DeviceExtension
  // will have { getDeviceInfo, doSomethingElse }
  DeviceExtension.getDeviceInfo().then(info => {
    console.log(info)
  })
}

})();
```

Prior to this change, the behavior depends on whether the above JS code
executes before or after Nimbus initialization.
With this change, both `getDeviceInfo` and `doSomethingElse` are
always available after initialization, regardless of initialization
order.

### DIFF
--- a/packages/nimbus-bridge/src/index.ts
+++ b/packages/nimbus-bridge/src/index.ts
@@ -30,7 +30,7 @@ class Nimbus {
       let extensionNames = JSON.parse(_nimbus.nativeExtensionNames());
       extensionNames.forEach((extension: string) => {
         Object.assign(window, {
-          [extension]: this.promisify(window[`_${extension}`])
+          [extension]: Object.assign(window[extension] || {}, this.promisify(window[`_${extension}`]))
         });
       });
     }
--- a/packages/test-www/test/nimbus-core-tests-setup.ts
+++ b/packages/test-www/test/nimbus-core-tests-setup.ts
@@ -1,0 +1,10 @@
+//
+// Copyright (c) 2020, Salesforce.com, inc.
+// All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+// For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+//
+
+const mochaTestBridge = window.mochaTestBridge || {}
+mochaTestBridge.myProp = "exists";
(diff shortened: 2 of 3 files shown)


packages/nimbus-bridge/src/index.ts
```diff
@@ -30,7 +30,7 @@ class Nimbus {
       let extensionNames = JSON.parse(_nimbus.nativeExtensionNames());
       extensionNames.forEach((extension: string) => {
         Object.assign(window, {
-          [extension]: this.promisify(window[`_${extension}`])
+          [extension]: Object.assign(window[extension] || {}, this.promisify(window[`_${extension}`]))
```
paultiarks: This seems okay to me. Only comment I would have would be that it would be nice, if it's possible, to add a test to verify this and keep it from regressing.

AdamLiechty: Done, thanks @paultiarks

AdamLiechty:
@threeve @sfdctaka the CLA is now approved and ready to merge if all still looks good. Thanks!

PR #124, opened 2020-01-24
Nimbus javascript Distribution RFC

paultiarks:
We don't currently have a good option for native consumers to retrieve a built, versioned copy of the nimbus js to inject from the native side. The need to inject nimbus from native seems like a valid use case, both for production and testing, and is something we should support for a 1.0 of Nimbus.

This RFC provides 3 slightly different options for iOS. I have a preference but I don't want to skew any opinions yet.

There is only a single option for Android, because I don't have much experience at all on Android and it seemed straightforward. This would be the area I would like feedback on the most, and any other options that I haven't thought of for Android would be appreciated.

### DIFF
--- a/docs/site/rfcs/nimbusJSDistribution.md
+++ b/docs/site/rfcs/nimbusJSDistribution.md
@@ -59,7 +59,7 @@ Cons:
 
 **1 - New Nimbus Module**
 
-We could add a new module to the Nimbus Android project. This module would contain a single Loader class that would vend the `nimbus.js` source. It would obtain the source with a phase added to the module's gradle script which would download the build artifact from the GitHub release, similar to the above iOS options.
+We could add a new module to the Nimbus Android project. This module would contain a single Loader class that would vend the `nimbus.js` source. At build time, the module would build the nimbus js source locally and include it as a resource in the resulting AAR. This module would also contain utilities to automatically inject the nimbus script into the `<head>` tag of a loading page. Ideally, this behavior would be the default, but also optional for consumers who wish to inject the nimbus js differently.
 
 Pros: 
 - Fairly simple, seems to be idiomatic for the platform
--- a/docs/site/rfcs/nimbusJSDistribution.md
+++ b/docs/site/rfcs/nimbusJSDistribution.md
@@ -10,7 +10,7 @@ This RFC discusses a number of options to enable this on both iOS and Android.
 
 # iOS
 
-**(1) Podspec with a `prepare_command` step**
+**1 - Podspec with a `prepare_command` step**
 
 ```ruby
 spec.prepare_command = <<-CMD
@@ -27,7 +27,7 @@ Cons:
 - This is not technically the intended use of `prepare_command`
 - Requires a non-trivial amount of effort to publish artifacts or automate releases (which we should probably do anyway)
 
-**(2) Podspec with a source specification**
+**2 - Podspec with a source specification**
 
 ```ruby
 spec.source = { :http => 'https://github.com/salesforce/nimbus/archive/0.0.7.zip' }
@@ -41,7 +41,7 @@ Pros:
 Cons: 
 - Requires a non-trivial amount of effort to publish artifacts or automate releases (which we should probably do anyway)
 
-**(3) Podspec with a script phase**
+**3 - Podspec with a script phase**
 
 ```ruby
 spec.script_phase = { :name => 'Hello World', :script => 'echo "Hello World"' }
@@ -57,7 +57,7 @@ Cons:
 
 # Android
 
-**(1) New Nimbus Module**
+**1 - New Nimbus Module**
 
 We could add a new module to the Nimbus Android project. This module would contain a single Loader class that would vend the `nimbus.js` source. It would obtain the source with a phase added to the module's gradle script which would download the build artifact from the GitHub release, similar to the above iOS options.
 
--- a/docs/site/rfcs/nimbusJSDistribution.md
+++ b/docs/site/rfcs/nimbusJSDistribution.md
@@ -2,7 +2,7 @@
 
 In order for nimbus to work, the `nimbus.js` file that is the build output of the TypeScript project has to be present as a resource in any webpage that wants to use hybrid features. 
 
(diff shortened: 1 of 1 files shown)


sfdctaka:
So this is for the case we want to include nimbus.js in native projects only? Should we also cover the case where if the user wants to include nimbus.js in the web app side? Or we should think this as a separate effort?

sfdctaka:
For iOS I like option 2 because it's the standard approach. I would have like option 1 more if it was standard because it won't require zipping and also because we can emphasize nimbus.js as a single  file to get. But this doesn't beat, in my mind, the standard approach in option 2.

I also leave Android consumption to @jamie-houston.

threeve:
+1 for option #2 on iOS. We should consider also package a source file in the pod that can easily be used to load the nimbus.js and inject it via `WKUserScript` to make it as automatic as possible.

docs/site/rfcs/nimbusJSDistribution.md
```diff
@@ -0,0 +1,68 @@
...
+
+Cons: 
+- Causes a warning in the `pod install` process. Invasive to consumers.
+- Need to rely on Xcode features to not run this build phase more often than needed.
+
+# Android
+
+**1 - New Nimbus Module**
+
+We could add a new module to the Nimbus Android project. This module would contain a single Loader class that would vend the `nimbus.js` source. It would obtain the source with a phase added to the module's gradle script which would download the build artifact from the GitHub release, similar to the above iOS options.
```
threeve: > It would obtain with a phase ... which would download the build artifact

I don't think this is the ideal way to do it. When the aar is built in order to publish the module to bintray it could simply build nimbus.js locally and package it as an asset.

The trick for android is how to inject the resource into the web app, because there is no equivalent to `WKUserScript`. I'm familiar with two possible approaches:
- Add an interceptor that will rewrite the `<head>` of the html document to reference `nimbus.js` via a `<script>` tag.
- Add a page load listener and load the nimbus.js after page load. 

The former has the benefit that it guarantees the nimbus runtime is available prior to any other scripts loading. However it feels somewhat hacky due to rewriting chunks of the html that is loaded. The capacitor framework uses this approach

The latter would require adding a "nimbus-ready" or similar event that can be published, and ensuring the contained web app does not attempt to use nimbus until the event has fired. This is the approach cordova uses.

Are there any other possibilities we can discover, @jamie-houston ?

docs/site/rfcs/nimbusJSDistribution.md
```diff
@@ -0,0 +1,68 @@
+# Distributing nimbus.js to native consumers
+
+In order for nimbus to work, the `nimbus.js` file that is the build output of the TypeScript project has to be present as a resource in any webpage that wants to use hybrid features. 
+
+The current, officially supported way that we do this is to add nimbus as an npm dependency on the web app and load `nimbus.js` in the page itself. This works in some cases, but there are many valid use cases that would be better suited to loading `nimbus.js` as a user script by a hybrid container itself.
```
wes566: I don't even like the npm option anymore... it feels like it's to easy to for the javascript and native sides to get out of sync version-wise.  We don't technically have the nimbus JS package up on NPM yet (so far in my apps I've been npm install'ing using the github repo link), but I'm not sure we should.  So I'm glad you are tackling this RFC. 💪 

Is the scope of this RFC just about how to deliver the JS file?  Or is it also about how to inject it into the webview so consumers don't have to modify their own HTML?

docs/site/rfcs/nimbusJSDistribution.md
wes566: +1 to packaging the JS as an asset in the AAR.

I know it's a little _magical_ but I prefer the first option you mention there about rewriting the `<head>`.  As a consumer of nimbus I think it's kind of nice if this stuff is just taken care of and I don't have to listen to some new event.  That said, if I wanted to get super customized with it I wouldn't mind there being some option (that you inject into the native constructor or something) that let's me opt-out of the magical rewriting of `<head>` and let me do that stuff manually.  The idea being that consuming nimbus is simple by default, but there is an escape hatch if I want to get all customize-y

docs/site/rfcs/nimbusJSDistribution.md
paultiarks: So far it has only been about how to deliver the JS. But I think they basically the same thing for native consumers, the more I think about it.

paultiarks:
Seems like we have quorum for Option 2 on iOS. I envisioned making this new pod a dependency in the nimbus podspec, and then we can create the extension/utility to inject the js in the main nimbus repo. Does that sound reasonable @threeve and @sfdctaka ?

docs/site/rfcs/nimbusJSDistribution.md
threeve: Yeah, I see them as more or less the same thing

paultiarks:
@threeve @wes566 @sfdctaka I incorporated the feedback so far, please have another look when you have a chance.

sfdctaka:
LGTM!

docs/site/rfcs/nimbusJSDistribution.md
```diff
@@ -0,0 +1,70 @@
...
+
+Cons: 
+- Causes a warning in the `pod install` process. Invasive to consumers.
+- Need to rely on Xcode features to not run this build phase more often than needed.
+
+# Android
+
+**1 - New Nimbus Module**
+
+We could add a new module to the Nimbus Android project. This module would contain a single Loader class that would vend the `nimbus.js` source. At build time, the module would build the nimbus js source locally and include it as a resource in the resulting AAR. This module would also contain utilities to automatically inject the nimbus script into the `<head>` tag of a loading page. Ideally, this behavior would be the default, but also optional for consumers who wish to inject the nimbus js differently.
```
jamie-houston: What does vend mean?

I think this sounds like a good solution.  Should be a simple thing to implement.  Famous last words....

paultiarks: Here I meant that it could supply the content of nimbus.js as a string, if the consumer wanted to not use the convenience utility we are going to provide and load the script themselves somehow.

PR #129, opened 2020-02-04
NimbusJS Podspec

paultiarks:
This creates the NimbusJS podspec which references the zip file published on the 0.0.7 nimbus release artifacts and is composed of the source found in that zip. This also includes a shell script that will assemble the necessary parts expected to be in the zip file. This script will be used primarily by the upcoming CI changes to publish release artifacts.

### DIFF
--- a/create-nimbus-js.sh
+++ b/create-nimbus-js.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+#remove old NimbusJS.zip if it exists
+rm NimbusJS.zip
+
+# compile nimbus.js
+cd packages/nimbus-bridge
+npm run build
+cd ../../
+
+# zip nimbus.js
+zip -j NimbusJS.zip packages/nimbus-bridge/dist/iife/nimbus.js
+
+# zip the webview extension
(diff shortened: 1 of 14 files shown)


create-nimbus-js.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
paultiarks: This is currently jus in the root of the project because I didn't know where to put it. Should this be moved into a `scripts` directory or something else?

platforms/apple/Sources/NimbusJS/WebView+NimbusJS.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+
+import WebKit
+
+enum NimbusJSError: Error {
+    case sourceNotFound
+}
+
+public extension WKWebView {
+    func injectNimbusJavascript(scriptName: String = "nimbus", bundle: Bundle? = nil) throws {
```
sfdctaka: Nice!  I kept having to inject nimbus.js in most of my mini sample apps so this is a time saver.

PR #130, opened 2020-02-04
Adding double underscores to emphasize nimbus as a pseudo private variable.

sfdctaka:


packages/nimbus-bridge/src/index.ts
```diff
@@ -211,13 +214,18 @@ class Nimbus {
...
+    __nimbus?: Nimbus;
   }
 }
-window.nimbus = nimbus;
 
-export default nimbus;
+if (window.__nimbus === undefined) {
+  window.__nimbus = __nimbus;
+} else {
+  window.__nimbus = Object.assign(window.__nimbus, __nimbus);
```
sfdctaka: Plugin might have been bounded prior to this line executing.  Instead of overwriting `window.__nimbus` merging the class with existing `__nimbus` on `window`.

packages/test-www/test/callback-encodable-tests.ts
```diff
@@ -15,54 +15,88 @@ interface MochaMessage {
...
+    completion: (param0: number, param1: number) => void
+  ): void;
+  callbackWithPrimitiveAndUddtParams(
+    completion: (param0: number, param1: MochaMessage) => void
+  ): void;
 }
 
-declare var callbackTestExtension: CallbackTestExtension;
+declare interface NimbusWithCallbackTestExtension {
+  callbackTestExtension: CallbackTestExtension;
```
sfdctaka: Typescript is very strict that there is `callbackTestExtension` object on nimbus. So I'm creating an interface here then casting to this interface at line 40.

platforms/apple/Sources/Nimbus/Connection.swift
```diff
@@ -59,12 +59,18 @@ public class Connection<C>: Binder {
     public func bind(_ callable: Callable, as name: String) {
         bindings[name] = callable
         let stubScript = """
-        \(namespace) = window.\(namespace) || {};
+        if (window.__nimbus === undefined) {
+            window.__nimbus = {};
+        }
+        if (window.__nimbus.\(namespace) === undefined) {
+            window.__nimbus.\(namespace) = {};
+        }
```
sfdctaka: Extensions are moved now as properties of `__nimbus`.  So first check existence of `__nimbus` and add namespaces accordingly.

platforms/apple/Sources/Nimbus/Connection.swift
```diff
@@ -59,12 +59,18 @@ public class Connection<C>: Binder {
     public func bind(_ callable: Callable, as name: String) {
         bindings[name] = callable
         let stubScript = """
-        \(namespace) = window.\(namespace) || {};
+        if (window.__nimbus === undefined) {
+            window.__nimbus = {};
```
threeve: I'm not a huge fan of this and would prefer we come up with a way of ensuring initialization happens in a well-defined order. Perhaps we can do something closer to what android does with a prefixed namespace that is then brought into nimbus when the nimbus runtime loads? This might be a candidate for an RFC and discussion.

sfdctaka: I will create an issue calling for RFC.  I agree, if there's a temporary place to stash extensions and then later merge that into `window.__nimbus.plugins` that would be much nicer than checking for namespace like this.

PR #133, opened 2020-02-12
Android Nimbus JS Distribution

paultiarks:
This does many things. A new module `nimbusjs` has been created in the Android project. In the gradle file for this module is executing a new bash script which builds nimbus, and copies the resulting nimbus.js into the resources of the new module. A utility class has been created that finds the `<head>` tag of an `InputStream` and injects nimbus.js. This will allow consumers to intercept page requests in their `WebViewClient` and use this utility function to inject nimbus.

All feedback is more than welcome, this is my first foray into the Android world and I expect there will need to be changes.


platforms/android/nimbusjs/src/androidTest/java/NimbusJSUtilsTests.kt
```diff
@@ -0,0 +1,62 @@
...
+        var containsNimbus = resultString?.contains("nimbus")
+        assertEquals(true, containsNimbus)
+    }
+
+    private fun readAssetStream(stream: InputStream): String? {
+        try {
+            val bufferSize = 1024
+            val buffer = CharArray(bufferSize)
+            val out = StringBuilder()
+            val `in`: Reader = InputStreamReader(stream, "UTF-8")
```
jamie-houston: nit: Preferably we shouldn't use keywords as variable names.

platforms/android/nimbusjs/src/main/java/com/salesforce/nimbusjs/NimbusJSUtilities.kt
```diff
@@ -0,0 +1,60 @@
...
+import android.webkit.WebView
+import java.io.*
+import java.nio.charset.StandardCharsets
+
+class NimbusJSUtilities() {
+    companion object Injection {
+
+        fun injectedNimbusStream(inputStream: InputStream, context: Context): InputStream {
+            val jsString = ResourceUtils(context).stringFromRawResource(R.raw.nimbus)
+            var html: String? = this.readAssetStream(inputStream)
```
jamie-houston: looks like this shouldn't have a null html... so maybe
```
var html = this.readAssetStream(inputStream) ?: throw MyAwesomeException()
```

platforms/android/nimbusjs/src/main/java/com/salesforce/nimbusjs/NimbusJSUtilities.kt
```diff
@@ -0,0 +1,60 @@
...
+            if (html!!.contains("<head>")) {
+                html = html.replace("<head>", "<head>\n$jsString\n")
+            } else if (html.contains("</head>")) {
```
jamie-houston: How does the html end up with either \<head\> or \</head\>?  Seems weird that it'd be missing the first but have the second...?

platforms/android/nimbusjs/src/main/java/com/salesforce/nimbusjs/NimbusJSUtilities.kt
```diff
@@ -0,0 +1,60 @@
...
+            }
+            return ByteArrayInputStream(html!!.toByteArray(StandardCharsets.UTF_8))
+        }
+
+        private fun readAssetStream(stream: InputStream): String? {
+            try {
+                val bufferSize = 1024
+                val buffer = CharArray(bufferSize)
+                val out = StringBuilder()
+                val `in`: Reader = InputStreamReader(stream, "UTF-8")
```
jamie-houston: nit: Use constant here too (`StandardCharsets.UTF_8`)

platforms/android/nimbusjs/src/main/java/com/salesforce/nimbusjs/NimbusJSUtilities.kt
paultiarks: In cases where `<head>` has attributes. We could probably shorten this to just checking for `</head>`

platforms/android/nimbusjs/src/androidTest/java/NimbusJSUtilsTests.kt
```diff
@@ -0,0 +1,62 @@
...
+    @Before
+    fun setup() {
+        context = InstrumentationRegistry.getInstrumentation().targetContext
+    }
+
+    @Test
+    fun testInjectionTwo() {
+        var inputStream: InputStream = context.assets.open("testPage.html")
+        inputStream = NimbusJSUtilities.injectedNimbusStream(inputStream, context)
+        var resultString = readAssetStream(inputStream)
```
jamie-houston: I think you can just use the helper `inputStream.bufferedReader().readText()` or something like that.

platforms/android/nimbusjs/src/main/java/com/salesforce/nimbusjs/NimbusJSUtilities.kt
jamie-houston: here too, I think you could just read it in with `inputStream.bufferedReader(StandardCharsets.UTF_8).readText()` and not create the (duplicated) helper method

jamie-houston: There's probably a way to filter it while you're reading it too, so if you wanted to just search for \<head\> and modify or copy to new string/stream you could.  But not sure the (perf) benefit of the added complexity.

platforms/android/nimbusjs/src/androidTest/java/NimbusJSUtilsTests.kt
paultiarks: Wow that cleans things up considerably. Thank you

platforms/android/nimbusjs/src/main/java/com/salesforce/nimbusjs/NimbusJSUtilities.kt
```diff
@@ -0,0 +1,60 @@
...
+            } catch (e: Exception) {
+                Log.e("Nimbus", "Unable to inject Nimbus", e)
+            }
+            return ""
+        }
+    }
+}
+
+private fun String.Companion.fromStream(stream: InputStream): String {
+    val result = ByteArrayOutputStream()
```
jamie-houston: Is this doing something different than the other stream to string functions?  Seems like the same thing...?

Also, I'm not sure of this paradigm... creating an extension method to String that is only used in this one method.  If you need this, I'd say you could make it a private function inside of the ResourceUtils class.

platforms/android/nimbusjs/src/main/java/com/salesforce/nimbusjs/NimbusJSUtilities.kt
jamie-houston: ahh... yeah that makes sense!

jamie-houston: Do you want to handle a case where there is no head section?  Not sure if that's valid html, but maybe you'd add the \<head\> if it was missing...? 🤷‍♂

platforms/android/nimbusjs/src/main/java/com/salesforce/nimbusjs/NimbusJSUtilities.kt
paultiarks: yep. good call, cleaning up now.

platforms/android/nimbusjs/src/main/java/com/salesforce/nimbusjs/NimbusJSUtilities.kt
paultiarks: I'm honestly not sure what to do in that case. I would lean away from adding a `<head>` if it's missing entirely. But I could be convinced.

jamie-houston: /shrug

platforms/android/nimbusjs/buildNimbusJS.sh
```diff
@@ -0,0 +1,13 @@
...
+
+cd ../../packages/nimbus-bridge/
+
+npm install
+
+npm run build
+
+mkdir ../../platforms/android/nimbusjs/src/main/res/raw/
+
+cp dist/iife/nimbus.js ../../platforms/android/nimbusjs/src/main/res/raw/nimbus.js
```
wes566: the android browser supports es modules so we should be using the ESM version of nimbus instead of iife, unless I'm missing something

paultiarks: If I remember correctly, there are older android versions that don't yet support it, so I used the iife version

PR #134, opened 2020-02-13
Update libraries to version 0.0.8

paultiarks:


threeve:
looks good, should we also sync the version in `packages/nimbus-bridge/package.json` ?

paultiarks:
Since the version in package.json is already `0.2.0` we should wait until we go 1.0 and get all 3 in sync at that point.

PR #138, opened 2020-02-24
Move global nimbus state to the module level

threeve:
- Use `Object.defineProperties` to make the global nimbus object's
properties non-configurable
- Hide state variables like promise and callback maps in the module
- Refactor iOS connection to stage plugin exports outside of the
nimbus object, which imports them on initialization


packages/nimbus-bridge/src/index.ts
```diff
@@ -8,227 +8,296 @@
...
-    [s: string]: { resolve: Function; reject: Function };
-  } = {};
-  private callbacks: { [s: string]: Function } = {};
+  // Called by native code to execute a pending callback function.
+  callCallback(callbackId: string, args: any[]): void;
 
-  // Dictionary to manage message&subscriber relationship.
-  public listenerMap: { [s: string]: Function[] } = {};
+  // Called by native code to execute a pending callback function.
+  callCallback2(callbackId: string, ...args: any[]): void;
```
paultiarks: is `callCallback2` intentional? 🤔

threeve: yes, there is both `callCallback` and `callCallback2` currently. In the near future those will be consolidated into just one method, the only difference is taking varargs arguments vs an array of arguments.

PR #144, opened 2020-03-03
Allow parameters to a Callable that conform to Decodable

paultiarks:
Also, change how parameters are passed from javascript. Previously on iOS, parameters were passed as objects. They're now passed as json strings which are attempted to be decoded if the parameter type is Decodable.

### DIFF
--- a/packages/nimbus-bridge/src/index.ts
+++ b/packages/nimbus-bridge/src/index.ts
@@ -141,6 +141,8 @@ let cloneArguments = (args: any[]): any[] => {
       } else {
         clonedArgs.push({ callbackId });
       }
+    } else if (typeof args[i] === "object") {
+      clonedArgs.push(JSON.stringify(args[i]))
     } else {
       clonedArgs.push(args[i]);
     }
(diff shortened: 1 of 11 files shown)


packages/nimbus-bridge/src/index.ts
```diff
@@ -141,6 +141,8 @@ let cloneArguments = (args: any[]): any[] => {
       } else {
         clonedArgs.push({ callbackId });
       }
+    } else if (typeof args[i] === "object") {
+      clonedArgs.push(JSON.stringify(args[i]))
```
threeve: 🎉

platforms/apple/Sources/Nimbus/DecodableUtils.swift
```diff
@@ -0,0 +1,29 @@
...
+//  Nimbus
+//
+//  Created by Paul Tiarks on 3/2/20.
+//  Copyright © 2020 Salesforce.com, inc. All rights reserved.
+//
+
+import Foundation
+
+private extension Decodable {
+    static func openedJSONDecode(using decoder: JSONDecoder, from data: Data) throws -> Self {
```
threeve: what does this name mean? to me it's not exactly obvious from the name what this does.

paultiarks: It's not the best name. It's meant to convey that the protocol itself is opened and providing a decoding implementation. We could change the name, or inline the implementation into `decodeJSON` since it's the only place this is called.

PR #147, opened 2020-03-05
Change how invoking promise-returning JavaScript functions works

threeve:
iOS:
- the `invoke` method is relocated from `Connection` to `NimbusBridge`
- Removed the `WebBinder` protocol

Android:
- Removed parameter to `@ExtensionMethod` annotation
- Reverted `ExtensionMethod` related changes in nimbus annotation processor
- Added `invoke` method to `NimbusBridge` with similar implementation to iOS

General:
- `invoke` supports calling any function accessible in the global
  scope, not just plugin methods
- `invoke` supports both functions that return Promises and those that
  return values
- removed the nimbus runtime method `callAwaiting` as this can act as
  a back door for unprivileged javascript to execute plugin methods


platforms/android/nimbus-compiler/src/main/java/com/salesforce/nimbus/NimbusProcessor.kt
```diff
@@ -114,320 +113,189 @@ class NimbusProcessor : AbstractProcessor() {
...
+
+                                if (it.asType().toString().equals("java.lang.String")) {
+                                    methodSpec.addParameter(String::class.java, it.simpleName.toString())
+                                } else if (it.asType().toString().startsWith("kotlin.jvm.functions.Function")) {
+                                    methodSpec.addParameter(String::class.java, it.simpleName.toString() + "Id", Modifier.FINAL)
+
+                                    val invoke = MethodSpec.methodBuilder("invoke")
+                                        .addAnnotation(Override::class.java)
+                                        .addModifiers(Modifier.PUBLIC)
+                                        // TODO: only Void return is supported in callbacks, emit an error if not void
```
jamie-houston: I know this was already a TODO, but curious if we should just go ahead and add this error or if there's a story to handle it.

platforms/android/nimbus-compiler/src/main/java/com/salesforce/nimbus/NimbusProcessor.kt
```diff
@@ -114,320 +113,189 @@ class NimbusProcessor : AbstractProcessor() {
...
+                                        it.asType(),
+                                        it.simpleName,
+                                        extClass,
+                                        "hashMapFromJSON",
+                                        it.simpleName,
+                                        TypeName.get(declaredType.typeArguments[0]),
+                                        TypeName.get(declaredType.typeArguments[1]))
                                 } else {
-                                    argBlock.add("arg$index != null ? new \$T(arg$index) : null,\n", ClassName.get("com.salesforce.nimbus", "PrimitiveJSONSerializable"))
+                                    // TODO: we also want to support kotlinx.serializable eventually
```
jamie-houston: Is this still valid?

platforms/android/nimbus/src/main/java/com/salesforce/nimbus/NimbusBridge.kt
```diff
@@ -58,6 +61,97 @@ class NimbusBridge {
         bridgeWebView = null
     }
 
+    private fun invokeInternal(
```
jamie-houston: Was there a reason this was broken out of Invoke?  Looks like it's only called from Invoke?

threeve: I originally had this version public as well, but made it private based on feedback from @paultiarks

platforms/android/nimbus/src/main/java/com/salesforce/nimbus/NimbusBridge.kt
```diff
@@ -58,6 +61,97 @@ class NimbusBridge {
...
+        identifierSegments: Array<String>,
+        args: Array<JSONSerializable?> = emptyArray(),
+        callback: ((String?, Any?) -> Unit)
+    ) {
+        val promiseId = UUID.randomUUID().toString()
+        promises[promiseId] = callback
+
+        val segmentArray = JSONArray(identifierSegments)
+        val segmentString = segmentArray.toString()
```
jamie-houston: Probably not a perf concern at all, but could/should we avoid the JSONArray object with something like 
```
val segmentString = indentifierSegments.joinToString(seperator=', ', prefix='[', postfix=']')
```
Otherwise segmentArray isn't used anywhere.  OTOH, it's easier to read it the way you have it... 🤷‍♂

platforms/android/nimbus/src/main/java/com/salesforce/nimbus/NimbusBridge.kt
```diff
@@ -58,6 +61,97 @@ class NimbusBridge {
...
+
+        val jsonArray = JSONArray()
+        args.forEachIndexed { _, jsonSerializable ->
```
jamie-houston: If you're ignoring the index, you could just do forEach

platforms/android/nimbus/src/main/java/com/salesforce/nimbus/NimbusBridge.kt
threeve: Using JSONArray ensures that strings are escaped properly which helps avoid any injection vulnerabilities.

PR #148, opened 2020-03-10
Rename podspec and bump versions

paultiarks:
This does three things

1. Rename Nimbus.podspec to NimbusBridge.podspec to fix a name collision
2. Unify all platform versions and update to 0.99.0 (as discussed offline with @threeve)
3. Add an explicit dependency between NimbusJS and NimbusBridge to ensure consumers aren't using incompatible versions of nimbus.js

We can likely wait to merge this until we are about to release 0.99.0

### DIFF
--- a/NimbusJS.podspec
+++ b/NimbusJS.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name            = 'NimbusJS'
-  s.version         = '0.0.9'
+  s.version         = '0.99.0'
   s.summary         = 'NimbusJS supplies the javascript necessary for the Nimbus framework'
   s.homepage        = 'https://github.com/salesforce/nimbus'
   s.source          = { :http => 'https://github.com/salesforce/nimbus/releases/download/0.0.9/NimbusJS.zip' }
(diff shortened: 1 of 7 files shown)


sfdctaka:
Curious. What is the significance behind 99?

threeve:
@sfdctaka 0.99 because it's getting close to but isn't quite 1.0 :-D

sfdctaka:
If we are to stay at 99 for a bit and need to do a release like we did for 0.0.10 can we still do it?  0.0.10 had to be created as one off from 0.0.9 so curious how 99 plays out in such scenario.

threeve:
0.999, 0.9999, 0.99999...   :-D 

Or 0.99.1, 0.99.2, 0.99.3

Plenty of options.

But the goal is to hit 1.0 soon once the last few remaining items are completed.

NimbusJS.podspec
```diff
@@ -12,4 +12,6 @@ Pod::Spec.new do |s|
   s.swift_version   = '4.2'
 
   s.ios.deployment_target = '11.0'
+
+  s.dependency 'NimbusBridge', '= 0.99.0'
```
sfdctaka: 🎉

PR #153, opened 2020-03-19
Unify version numbers across platforms

paultiarks:
This change rallies each platforms version around a single version number, ensuring that version numbers move forward uniformly and with minimal effort.

The single source of version number will be the `lerna.json` file in the project root. We are already using lerna to unify the version of the typescript packages, so this seems like a reasonable resource to make as the source of truth.

When a new version needs to be created, we can run the `bump` task which will move the version number forward.

Android and iOS will then refer to the `lerna.json` file in their respective build scripts and podspecs.

Absent from this PR is any sort of CI check to make sure that version numbers are, in fact, in sync and nothing has broken with the various implementations. Would be interested in feedback on whether that kind of check would be necessary.

### DIFF
--- a/platforms/apple/Nimbus.xcodeproj/project.pbxproj
+++ b/platforms/apple/Nimbus.xcodeproj/project.pbxproj
@@ -587,6 +587,7 @@
 			isa = PBXNativeTarget;
 			buildConfigurationList = 291A860C21E530BC00510832 /* Build configuration list for PBXNativeTarget "Nimbus" */;
 			buildPhases = (
+				CC13116724256DBA002D698D /* Set Version */,
 				29CD07F62214BC9D007D3AF3 /* Run SwiftLint */,
 				291A85FE21E530BC00510832 /* Headers */,
 				291A85FF21E530BC00510832 /* Sources */,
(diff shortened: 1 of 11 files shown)


paultiarks:
Yep I forgot all about Carthage. Will figure that out now

paultiarks:
The latest change appends some tasks to the `bump` task. It executes a new local bash script which parses out the lerna version, and then sets it in the Nimbus and NimbusJS plists.

paultiarks:
The latest change undoes all of the changes from the last change and moves the version setting to script phases in the Nimbus project file.

platforms/apple/Nimbus.xcodeproj/project.pbxproj
```diff
@@ -984,6 +986,42 @@
...
+			inputPaths = (
+			);
+			name = "Set Version";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "FULL_VERSION=\"$(grep -Ei '\\\"version\\\": \\\"([.0-9]+)\\\"' ../../lerna.json)\"\nVERSION=\"$(echo ${FULL_VERSION}\\ | grep -E -o '[0-9]+.[0-9]+.[0-9]')\"\n\n/usr/libexec/PlistBuddy -c \"Set :CFBundleShortVersionString ${VERSION}\" $BUILT_PRODUCTS_DIR/$INFOPLIST_PATH\n";
```
sfdctaka: ✔️

PR #160, opened 2020-03-24
Documentation Updates

paultiarks:
This includes some minor corrections, instructions for injecting nimbus.js for both platforms, and a basic guide for writing a plugin on Android. Would especially like review for Android related items.

### DIFF
--- a/docs/site/docs/ios/writing-a-plugin.md
+++ b/docs/site/docs/ios/writing-a-plugin.md
@@ -48,7 +48,7 @@ dharmaPlugin.checkBoat("Penny", boatName => {
 });
 ```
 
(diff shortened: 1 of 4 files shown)


docs/site/docs/ios/writing-a-plugin.md
```diff
@@ -48,7 +48,7 @@ dharmaPlugin.checkBoat("Penny", boatName => {
 });
 ```
 
-Any native parameters or return types must conform to `Encodable` in order for them to be bound. You can bind methods with up to 5 parameters, the last of which can be a callback. If you need more parameters than is available, it's recommended that you build an `Encodable` struct to wrap the necessary context.
+Any native parameters or return types must conform to `Decodable` in order for them to be bound. You can bind methods with up to 5 parameters, the last of which can be a callback. If you need more parameters than is available, it's recommended that you build an `Decodable` struct to wrap the necessary context.
```
threeve: parameter types-> `Decodable`
return types -> `Encodable`

PR #161, opened 2020-03-25
Execute callback invocation in a scope to avoid global pollution

### LINKED ISSUE #68: Memory leak in Callback.swift
Swift callback (and potentially android as well) are accidentally leaking data into global variables. Callbacks and other javascript invocations from native should wrap in an anonymous function invocation so we don't leak into the global scope.

threeve:
fixes #68

### DIFF
--- a/platforms/apple/Sources/Nimbus/Callback.swift
+++ b/platforms/apple/Sources/Nimbus/Callback.swift
@@ -50,6 +50,7 @@ class Callback: Callable {
 
         DispatchQueue.main.async {
             self.webView?.evaluateJavaScript("""
+            {
                 var jsonArgs = \(formattedJsonArgs);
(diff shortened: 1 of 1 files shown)


platforms/apple/Sources/Nimbus/Callback.swift
```diff
@@ -50,6 +50,7 @@ class Callback: Callable {
 
         DispatchQueue.main.async {
             self.webView?.evaluateJavaScript("""
+            {
                 var jsonArgs = \(formattedJsonArgs);
```
sfdctaka: Since we enclosed the code in a block wondering if we should also change `var` to `const`?

threeve: I don't think that's necessary in this case.

PR #162, opened 2020-03-26
Update docs

threeve:


paultiarks:
It looks like there are a couple of pages created that don't yet have links in the navigation of the docs. Do you want to add those here or should I make another PR with them?

docs/site/_includes/docs.njk
```diff
@@ -60,10 +55,15 @@ layout: base
                             >Installation</a
                         >
                     </li>
+                    <li>
+                        <a href="{{ '/docs/android/writing-a-plugin' | url }}"
+                            >Writing a Plugin</a
```
paultiarks: 👍

PR #163, opened 2020-03-27
Refactoring & cleanup in preparation for headless bindings

mattkranzler5:
First of all I apologize in advance for the size of this PR. Due to the nature of the refactoring I felt it made sense to do it in one fell swoop. 

The purpose of this PR is to get the Android modules/project into a state to allow adding support for headless (v8) bindings. Along the way I also cleaned some things up (fixed various warnings, updated some dependency versions, etc.). I will do my best to describe the changes and why I feel they are necessary. 

### Changes
- Created a `nimbus-core` Kotlin module which contains base interfaces and other shared code between various nimbus bridge modules. Moved base interfaces from the original `nimbus` module to this new module
- Extracted a base `Bridge` interface from the original `Bridge` class and added a `JavascriptEngine` type (`WebView`, `V8`, etc.)
- Added a `JavascriptEngine` type to the `Binder` interface to allow for separate binders for each `JavascriptEngine`
- Created a `Runtime` interface which removes a circular dependency between the `Plugin` and the `Bridge`. This allows us to provide the `Plugin` a way of doing some customization and cleanup on the `JavascriptEngine` and also `invoke()` a function on the `JavascriptEngine` that passes through the `Bridge` without exposing the `Bridge` api to the `Plugin` directly.
- Renamed the `nimbus` module to `nimbus-bridge-webview` and the `Bridge` implementation to `WebViewBridge`
- Renamed the `nimbus-compiler` module to `nimbus-compiler-webview` and the `NimbusProcessor` class to `WebViewBinderGenerator` to be more indicative of it's responsibility
- Removed `appcompat` dependency from `nimbus-bridge-webview` (formerly `nimbus`) module since it was only used in an instrumentation test and doesn't need to be included

### Justification of Changes
The refactoring of the modules and interfaces allows us to easily add new bridges to different Javascript Engines, such as v8. To do so we will just need to create a new module for that bridge and then implement the `Bridge` interface, as well as create a new module for the compiler to generate the `Binder`s for that `Bridge`. This architecture also allows consumers to use one or many different `Bridge`s  in their project as well as a single `Plugin` with the ability to customize for each Javascript Engine. 

This PR will also create a new artifact (`nimbus-bridge-webview`) which will replace the `nimbus` artifact. Since there are some pretty big api breaking changes and this code will go into the 2.0 release I felt it made more sense to create a separate artifact to make the upgrade more explicit.


platforms/android/nimbus-core/src/main/java/com/salesforce/nimbus/Runtime.kt
```diff
@@ -0,0 +1,21 @@
...
+     * Get the [JavascriptEngine] powering the [Runtime].
+     */
+    fun getJavascriptEngine(): JavascriptEngine?
+
+    /**
+     * Invokes a [functionName] in the [JavascriptEngine].
+     */
+    fun invoke(
+        functionName: String,
+        args: Array<JSONSerializable?> = emptyArray(),
```
jhorst11: does this assume that the args need to be serialized to JSON? Is that an implementation detail since certain runtimes (i.e. V8) won't require the args to be serialized.

threeve: I believe the constraint that the argument is representable as JSON is still necessary, even if the arguments are never actually serialized to a string. For primitive types they can still be passed directly, but objects may get mapped directly into a V8 object instead of round-tripping through a string representation.

platforms/android/build.gradle
```diff
@@ -1,12 +1,12 @@
 buildscript {
-    ext.kotlin_version = '1.3.50'
+    ext.kotlin_version = '1.3.61'
```
jamie-houston: Should we just bump to 1.3.70?

platforms/android/nimbus-annotations/build.gradle
```diff
@@ -6,14 +8,12 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
 }
 
-sourceCompatibility = "7"
-targetCompatibility = "7"
+sourceCompatibility = "8"
```
jamie-houston: I know you're just updating what's already in here, but I've only ever seen or set this as "1.8" and not "8". And always inside of an `android` or 'java' block, never at the top level.  Does declaring it here just apply it to everything?

Also, we could reduce some redundancy across all the build.gradle files by setting this in allProjects in the root build.gradle.  Or using buildSrc.

mattkranzler5: Yeah, I'll move this to the root file. At some point it would be good to migrate to kotlin build files but I didn't want to do that as part of this PR.

mattkranzler5: @jamie-houston turns out those statements aren't used so I deleted them.

PR #167, opened 2020-04-02
Refactor for JSCore Binding

paultiarks:
This is a first step in refactoring to allow for binding to a JSContext. This creates a JSContext Bridge class and connection. To support that, Connection is now a protocol implemented for each binding environment.

Most of the JSContext classes are largely unimplemented as of yet, as this is a check-in merge to prevent the JSCore work from diverging too much.


platforms/apple/Sources/Nimbus/Plugin.swift
```diff
@@ -19,5 +18,11 @@ public protocol Plugin: class {
      to the web view, make additional configuration changes to the web view, or
      call additional methods on the nimbus bridge prior to the web app being loaded.
      */
-    func bind(to webView: WKWebView, bridge: Bridge)
+    func bind<C>(to connection: C) where C: Connection
+}
+
+public extension Plugin {
+    var namespace: String {
+        return String(describing: self)
```
threeve, line 26: Have you tested this? I think this may by default include the module name of the type, and for nested types may include the containing type name as well?

platforms/apple/Sources/Nimbus/Plugin/DeviceInfoPlugin.swift
```diff
@@ -53,7 +53,7 @@ struct DeviceInfo: Codable {
     #endif
 }
 
-public class DeviceInfoPlugin {
+public final class DeviceInfoPlugin {
```
threeve, line 56: `final` shouldn't be necessary any more, is this leftover from some of the attempts we made adding a `Self` type constraint or is it intentional?

platforms/apple/Sources/Nimbus/JSContext+Nimbus.swift
```diff
@@ -0,0 +1,15 @@
...
+//  JSContext+Nimbus.swift
+//  Nimbus
+//
+//  Created by Paul Tiarks on 3/31/20.
+//  Copyright © 2020 Salesforce.com, inc. All rights reserved.
+//
+
+import JavaScriptCore
+
+extension JSContext {
```
threeve, line 11: I would prefer not to have this extension and just create a `JSContextConnection` directly inside the `JSContextBridge`

platforms/apple/Sources/Nimbus/JSContextConnection.swift
```diff
@@ -0,0 +1,31 @@
...
+// Copyright (c) 2019, Salesforce.com, inc.
+// All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+// For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+//
+
+import JavaScriptCore
+
+public class JSContextConnection<C>: Connection {
+    public typealias Target = C
```
threeve, line 11: this type alias is not used

platforms/apple/Sources/Nimbus/WebView+Nimbus.swift
```diff
@@ -11,8 +11,8 @@ extension WKWebView {
     /**
      Create a connection from this web view to the specified object.
      */
-    public func addConnection<C>(to target: C, as namespace: String) -> Connection<C> {
-        return Connection(from: self, to: target, as: namespace)
+    public func addConnection<C>(to target: C, as namespace: String) -> WebViewConnection<C> {
```
threeve, line 14: Can we remove this extension method and just create a `WebViewConnection` directly from `WebViewBridge`?

platforms/apple/Sources/Nimbus/WebViewConnection.swift
```diff
@@ -0,0 +1,156 @@
...
+import WebKit
+
+/**
+ A `Connection` links a web view to native functions.
+
+ Each connection can bind multiple functions and expose them under
+ a single namespace in JavaScript.
+ */
+public class WebViewConnection<C>: Connection {
+    public typealias Target = C
```
threeve, line 17: this type alias is not used

platforms/apple/Sources/Nimbus/JSContextBridge.swift
```diff
@@ -0,0 +1,19 @@
...
+// Copyright (c) 2019, Salesforce.com, inc.
+// All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+// For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+//
+
+import Foundation
+import JavaScriptCore
+
+public class JSContextBridge: NSObject {
```
threeve, line 11: is inheriting `NSObject` necessary here?

platforms/apple/Sources/Nimbus/Plugin.swift
paultiarks: The DeviceInfoPlugin is using it, and those tests pass. But I'll double check the nested type case.

platforms/apple/Sources/Nimbus/Plugin/DeviceInfoPlugin.swift
paultiarks: Yep, it likely is. Will remove.

PR #168, opened 2020-04-07
Now generating Kotlin Binders, removed "nimbus" module prefix

mattkranzler5:
This PR generates Kotlin Binders using kotlin-poet instead of Java Binders using java-poet. It also pulls in the kotlin-metadata library in order to support nullable parameters & return types in the Binders. 

### Changes:
- Generated Binders are in Kotlin instead of Java
- Created new `compiler-base` library which has an abstract processor which each compiler module will include
- Cleaned up the compiler/processor to make things more readable and allow for subclasses to provide specific functionality for the javascript engine it is targeting
- Removed the `nimbus-` prefix from module names since our group name is already `com.salesforce.nimbus` to remove redundancy
- Added support for a typed key in the `PrimitiveExtensions.hashMapFromJSON()` function


mattkranzler5:
👍

PR #173, opened 2020-04-07
Remove support for `NSArray` and `NSDictionary` in binders

threeve:
Support for `NSArray` and `NSDictionary` types was added to support a POC but is not actually used. Removing it _significantly_ reduces the complexity of `Binder` by eliminating probably 3/4 of the total overloads. Starting with 2.0 we will explicitly require the use of `Decodable` and `Encodable` only and not support foundation types that use `JSONSerialization`


sfdctaka:
🙌

PR #177, opened 2020-04-10
JSValueEncoder/Decoder

paultiarks:
The main change in this is the addition of JSValueDecoder (adapted from an old example from @threeve) and JSValueEncoder which is heavily influenced by JSONEncoder.

This also includes some changes in JSContextBridge and JSContextConnection with regard to how bound plugin functions are exposed to JS. I don't expect these changes to be final, and will likely be revised in the next phase of JSCore binding work.


platforms/apple/Sources/Nimbus/JSValueEncoder.swift
```diff
@@ -0,0 +1,560 @@
+//
+//  JSValueEncoder.swift
+//  Nimbus
+//
+//  Created by Paul Tiarks on 4/7/20.
```
threeve, line 5: need to put the OSS header here instead of Xcode's default

PR #178, opened 2020-04-10
Basic kotlinx-serialization support

mattkranzler5:
This adds basic kotlinx-serialization support along side the existing `JSONSerializable` way of serializing to/from a `WebView`. This allows you to just annotate a class with the Kotlin annotation `@Serializable` and in most basic cases the serialization & deserialization will just work. More advanced cases with generic types are currently not supported due to an open [bug](https://github.com/Kotlin/kotlinx.serialization/issues/685) where generic serialization is broken when using `kapt`. More advanced cases can be added later or can currently be accomplished using the existing `JSONSerializable` serialization.

### Changes
- Added kotlinx-serialization to the project
- Created new `JavascriptSerializable` interface which takes a `SerializedOutputType` type representing the type that the output is serialized to (`String` in the case of a `WebView`)
- `JSONSerializable` extends `JavascriptSerializable` to serialize to a `String`
- Added a `KotlinJSONSerializable` which wraps a kotlin `@Serializable` value and serializes using the kotlin serializer
- Updated `MochaTests` to add a couple of serializable parameters
- Updated Kotlin to `1.3.71` and Android Gradle Plugin to `3.6.2` (latest stable)

### DIFF
--- a/platforms/android/compiler-base/src/main/java/com/salesforce/nimbus/compiler/CompilerExtensions.kt
+++ b/platforms/android/compiler-base/src/main/java/com/salesforce/nimbus/compiler/CompilerExtensions.kt
@@ -79,7 +79,7 @@ fun TypeMirror.asRawTypeName(nullable: Boolean = false): TypeName {
 /**
  * Converts a Java [TypeName] from the [Element] to a Kotlin [TypeName]
  */
-fun Element.asKotlinType(nullable: Boolean = false) = asType().asKotlinTypeName(nullable = nullable)
+fun Element.asKotlinTypeName(nullable: Boolean = false) = asType().asKotlinTypeName(nullable = nullable)
 
 /**
  * Converts an [Element] to a [TypeName]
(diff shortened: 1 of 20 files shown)


platforms/android/core/build.gradle
```diff
@@ -36,8 +37,9 @@ android {
 }
 
 dependencies {
-    api project(":annotations")
+    implementation project(":annotations")
```
threeve, line 40: Will this mean that a consumer of nimbus now has to explicitly add the nimbus-annotations module as a dependency in their project?

mattkranzler5: No, it's exported in the `bridge-webview` module. This is to avoid duplicate declarations resulting in more than one class ending up in the class path.

platforms/android/compiler-base/src/main/java/com/salesforce/nimbus/compiler/CompilerExtensions.kt
```diff
@@ -64,11 +64,40 @@ fun TypeName.toKotlinTypeName(nullable: Boolean = false): TypeName {
  */
 fun TypeMirror.asKotlinTypeName(nullable: Boolean = false) = asTypeName().toKotlinTypeName(nullable = nullable)
 
+/**
```
jamie-houston, line 67: nit: There's a mix between `asKotlinType`, `asKotlinTypeName`, and `toKotlinTypeName`.  The last is to/from TypeName so that could be the reason.  But should `asKotlinType` be `asKotlinTypeName`?

mattkranzler5: Good catch, I'll update this.

PR #183, opened 2020-04-15
JSContextConnection bound function handling

paultiarks:
This adds handling for bindable function types to JSContextConnection as well as tests

### DIFF
--- a/platforms/apple/Sources/Nimbus/JSValueCallback.swift
+++ b/platforms/apple/Sources/Nimbus/JSValueCallback.swift
@@ -1,0 +1,35 @@
+// Copyright (c) 2020, Salesforce.com, inc.
+// All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+// For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+//
+
+import Foundation
+import JavaScriptCore
+
+class JSValueCallback: Callable {
+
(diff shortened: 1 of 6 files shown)


platforms/apple/Sources/Nimbus/JSValueCallback.swift
```diff
@@ -0,0 +1,35 @@
...
+
+class JSValueCallback: Callable {
+
+    init(callback: JSValue) {
+        self.callback = callback
+    }
+
+    func call(args: [Any]) throws -> Any {
+        guard let context = callback?.context else {
+            return 0
```
threeve, line 18: seems like throwing an error would be more appropriate here maybe?

platforms/apple/Sources/Nimbus/JSValueCallback.swift
```diff
@@ -0,0 +1,35 @@
...
+            if type(of: arg) as? Encodable.Type != nil {
+                let encodableArg = arg as! Encodable // swiftlint:disable:this force_cast
+                return try encodableArg.toJSValue(context: context)
+            } else {
+                throw ParameterError.conversion
+            }
+        }
+
+        // call the function
+        return callback?.call(withArguments: jsArgs) ?? 0
```
threeve, line 31: same here, if `call(withArguments:)` returns nil we should probably throw, rather than return 0

PR #184, opened 2020-04-15
Kotlin serialization to and from V8 (k2v8)

mattkranzler5:
This PR creates a new module (`k2v8`) which adds serialization support to serialize kotlin classes to and from V8Objects. It currently supports all primitive types and those that are supported by V8. It is manly intended to be used with value objects and classes that have functions or other non-primitive or non-serializable classes are not supported.


platforms/android/k2v8/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
jamie-houston, line 1: It looks like this is already covered in the [android root .gitignore](https://github.com/salesforce/nimbus/blob/master/platforms/android/.gitignore#L98)...?

PR #186, opened 2020-04-20
Call javascript functions from plugins

paultiarks:
This adds support for JSContextBridge to call javascript and return the result to the caller.

### DIFF
--- a/platforms/apple/Sources/Nimbus/JSContextConnection.swift
+++ b/platforms/apple/Sources/Nimbus/JSContextConnection.swift
@@ -53,7 +53,11 @@ public class JSContextConnection: Connection {
         connectionValue?.setObject(binding, forKeyedSubscript: name)
     }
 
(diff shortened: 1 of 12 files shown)


platforms/apple/Sources/Nimbus/JSContextConnection.swift
```diff
@@ -9,8 +9,9 @@ import JavaScriptCore
 
 public class JSContextConnection: Connection {
 
-    init(from context: JSContext, as namespace: String) {
+    init(from context: JSContext, bridge: JSContextBridge, as namespace: String) {
```
paultiarks, line 12: One thought I had but wasn't sure about was that instead of specifically typing for each bridge in the connection constructors, we could create a protocol with the `invoke` function. Connection could then implement this, as well as each respective bridge and this parameter could be typed as that. Thoughts?

threeve: I like this idea

paultiarks: Updated, can you have another look?

PR #191, opened 2020-04-22
Fix bintray upload

paultiarks:
Same changes from master but for the release 1.x branch

### DIFF
--- a/platforms/android/gradle/publishing.gradle
+++ b/platforms/android/gradle/publishing.gradle
@@ -52,6 +52,7 @@ bintray {
     def versionFile = file('../../../lerna.json')
     def parsedVersion = new JsonSlurper().parseText(versionFile.text)
     def versionToPublish = parsedVersion.version.toString()
+    project.version = versionToPublish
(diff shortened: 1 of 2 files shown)


platforms/android/gradle/publishing.gradle
```diff
@@ -47,10 +49,15 @@ install {
 
 
 bintray {
+    def versionFile = file('../../../lerna.json')
+    def parsedVersion = new JsonSlurper().parseText(versionFile.text)
+    def versionToPublish = parsedVersion.version.toString()
+    project.version = versionToPublish
```
jamie-houston, line 55: Should the version be set outside of this block so it's consistent across this file?  Looks like right now it's set to the value in the gradle.properties file in https://github.com/salesforce/nimbus/pull/191/files#diff-8a7b50dee3b66600a9178d59325486a8R14 but then changed here to the value in the lerna file.  We should be able to change the above line from 

` version = VERSION `

to

` version = versionToPublish `
(and add the parsing above that....)

and then keep the project.version the same below.

paultiarks: Yep, that seems to work. Updated

PR #192, opened 2020-04-23
Add documentation to new JSCore API

paultiarks:
This also changes the name of a couple of classes for consistency.

### DIFF
--- a/docs/site/docs/ios/plugin-testing.md
+++ b/docs/site/docs/ios/plugin-testing.md
@@ -39,7 +39,7 @@ class DharmaPluginTests: NimbusPluginTestsBase {
         let jsExpectation = expectation(description: "js")
         testBridge.currentExpectation = jsExpectation
         let js = """
-                dharmaPlugin.theNumbers().then(function(result) {
+                __nimbus.plugins.dharmaPlugin.theNumbers().then(function(result) {
                         testBridge.receiveResult(result);
                     });
                 """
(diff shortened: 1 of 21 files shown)


docs/site/docs/ios/plugin-testing.md
```diff
@@ -103,8 +103,8 @@ class NimbusPluginTestsBase: XCTestCase, WKNavigationDelegate {
 
         if let nimbusScript = try? String(contentsOfFile: nimbusPath) {
             let userScript = WKUserScript(source: nimbusScript, injectionTime: .atDocumentStart, forMainFrameOnly: true)
-            webView.configuration.userContentController.addUserScript(userScript)
             nimbus.attach(to: webView)
+            webView.configuration.userContentController.addUserScript(userScript)
```
threeve, line 107: is this right? I thought we said we needed to inject the nimbus runtime before attaching it to the webview?

platforms/apple/Sources/Nimbus/JSValueCallback.swift
```diff
@@ -9,11 +9,18 @@
 import Foundation
 import JavaScriptCore
 
+/**
+
```
threeve, line 13: 😐

docs/site/docs/ios/plugin-testing.md
paultiarks: this is actually correct. We need to attach first so that the bridge will cause the plugins to add their plugin exports to the web view, then injecting the runtime script to grab the exports. The other order causes the exports to never be imported.

platforms/apple/Sources/Nimbus/JSValueCallback.swift
paultiarks: woops, this is actually an accident. I'll fix

PR #193, opened 2020-04-27
Refactor bindings implementation

threeve:


packages/nimbus-bridge/src/index.ts
```diff
@@ -114,19 +114,7 @@ let cloneArguments = (args: any[]): any[] => {
...
-        _nimbus.makeCallback !== undefined
-      ) {
-        // TODO: Android passes only the callbackId string, whereas iOS passes an
-        // object with the callbackId property. These need to be merged and handled
-        // the same way to eliminate extraneous code paths
-        clonedArgs.push(callbackId);
-      } else {
-        clonedArgs.push({ callbackId });
-      }
+      clonedArgs.push(callbackId);
```
threeve, line 117: android and ios now pass callbacks across their bridge in the exact same way, just a string uuid.

platforms/apple/Sources/Nimbus/CallableBinder.swift
```diff
@@ -6,239 +6,343 @@
...
 /**
  `CallableBinder` is an extension to `Binder` that provides default
  implementations of all `bind` overloads based around the use of the
  `Callable` implementations to wrap function arguments.
  */
-protocol CallableBinder: Binder {
-    func bind(_ callable: Callable, as name: String)
+protocol CallableBinder: class, Binder {
+    // Bind a generic `Callable` into this specific bridge
+    func bindCallable(_ name: String, to callable: @escaping Callable)
```
threeve, line 27: this used to be just `bind`, but now that `Callable` is simply a type alias for a function that lead to an ambiguity, so this version had to be renamed.

platforms/apple/Sources/Nimbus/CallableBinder.swift
```diff
@@ -6,239 +6,343 @@
...
     public func bind<A0>(
-        _ function: @escaping (A0) throws -> Void,
-        as name: String
+        _ name: String,
+        to function: @escaping (A0) throws -> Void
     ) where A0: Decodable {
-        let callable = make_callable(function)
-        bind(callable, as: name)
+        bindCallable(name) { [weak self] (args: [Any?]) in
+            guard let self = self else { throw DecodeError() }
```
threeve, line 74: basically all the bind overloads now follow this pattern:

1. get strong ref to self
2. check the args count
3. decode each arg, either as an argument value or callback as appropriate
4. call the original function with the decoded args
5. encode the return value of the original function

PR #194, opened 2020-04-29
[iOS] Fix native-->JS Bridge.invoke()

AdamLiechty:
1. Passing arguments to the JS function was broken.
   rawArgs.map(a => { a.v }) // The passed function does not return anything
2. invoke should have JS functions specified like 'my.namespace.myFn'
   to match Android, rather than 'my,namespace,myFn' (commas)

One of the tests is updated to exercise passing of arguments to
a function that is not directly on the window object.


threeve:
Thanks Adam, I think we fixed this on master and forgot to backport the fix to 1.x. 👍

PR #197, opened 2020-05-01
V8 (headless) bridge support

mattkranzler5:
This PR adds a couple new modules, `bridge-v8` and `compiler-v8`. The `bridge-v8` module includes the `V8Bridge` which allows plugins to bind to a `V8` runtime. The `compiler-v8` module adds a code generator to generate plugin `Binder` classes that target `V8`. There were also some other changes:

- Cleaned up dependencies
- Renamed `JavascriptSerializable` to `JSEncodable` along with the `serialize()` function being renamed to `encode()`
- Much of the code in the `BinderGenerator` abstract class was moved to `WebViewBinderGenerator` since it couldn't be shared with the `V8BinderGenerator`


platforms/android/bridge-v8/src/androidTest/java/com/salesforce/nimbus/bridge/v8/PrimitiveV8EncodableTests.kt
```diff
@@ -0,0 +1,87 @@
...
+                try {
+                    a.toV8Encodable(v8)
+                } catch (e: Exception) {
+                    sameExceptionMessage = e.message.equals("Double value should be finite.")
+                }
+                sameExceptionMessage
+            } else {
+                val array = a.toV8Encodable(v8).encode() as V8Array
+                val value = array.getDouble(0)
+                a == value
```
jamie-houston, line 45: Is this the same thing as `a shouldBe value`? I'd prefer in tests to be specific about assertions if so.  But could be convinced otherwise.

jamie-houston:
Not necessarily in this PR, but we could add buildSrc to reduce the build.gradle and version redundency.  I used buildSrcVersions (which is apparently called [refreshVersions](https://github.com/jmfayard/refreshVersions) now)

platforms/android/bridge-v8/src/androidTest/java/com/salesforce/nimbus/bridge/v8/PrimitiveV8EncodableTests.kt
mattkranzler5: @jamie-houston I do like that more however due to `forAll()` requiring a boolean return (and `shouldBe` returns `Unit`) that won't work in this case.

jamie-houston: Ahh... looks like forAll with Gen requires boolean return but without allows assertions... didn't realize that.

PR #200, opened 2020-05-04
Refactored compilers to be a single module

mattkranzler5:
KAPT does not support multi-round annotation processing meaning that we can't have two processors consume the same annotation, therefore one wouldn't be able to have both V8 & WebView binders generated with two different compilers. This PR adds an interface (`BinderGenerator`) which is implemented for both V8 & WebView and moves the code from the generators to these new objects. There is now a single `compiler` module where the `PluginProcessor` delegates some code gen to each `BinderGenerator`, depending on if the `Bridge` class for the `BinderGenerator` exists in the classpath. This actually turns out better because the user only needs to add one `compiler` dependency and we can generate one or both Binders depending on if they also have either bridge dependency.


mattkranzler5:
Closing this due to it not working as expected.

PR #202, opened 2020-05-12
Added shared tests to verify plugins against v8 & webview js engines

mattkranzler5:
This PR adds a `shared-tests.js` file to the `test-www` package which includes a bunch of vanilla javascript functions to test various plugin functions. There is a new `shared-tests` Android module which includes both the `bridge-v8` and `bridge-webview` modules and runs the shared tests against both engines. This will help us find any compatibility issues between javascript engines and the plugins.


platforms/android/shared-tests/src/androidTest/assets/index.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
```
paultiarks, line 1: Did you explore any way to make this file shared between platforms?

platforms/android/shared-tests/src/androidTest/assets/webview-init.js
```diff
@@ -0,0 +1,3 @@
+window.onload = () => {
```
paultiarks, line 1: Can this be moved to a script tag in index.html?

platforms/android/shared-tests/src/androidTest/assets/index.html
mattkranzler5: I wasn't sure if iOS would need this file. If so, I can move it to the shared test directory. Does iOS need it as well @paultiarks?

platforms/android/shared-tests/src/androidTest/assets/webview-init.js
mattkranzler5: Yeah, it can. I'll move it.

platforms/android/shared-tests/src/androidTest/assets/index.html
paultiarks: iOS will definitely need something to load in the webview to exercise these tests. Having it shared (similar to how the mocha tests html is shared) seems like it would be preferable, unless there is some technical barrier that would make that impossible.

mattkranzler5: Ok, I'll see what I can do.

.circleci/config.yml
```diff
@@ -184,6 +184,14 @@ jobs:
     docker:
       - image: circleci/android:api-29-node
 
+    resource_class: medium+
+
+    environment:
+
+      # Configure the JVM and Gradle to avoid OOM errors
```
threeve, line 191: 🎉

Issue #111, opened 2020-01-02
Update plugin api to avoid initialization order issues


threeve:
It is possible to add plugins after attaching the bridge to the webview, however attaching the webview needs to happen after all plugins are added to the bridge. The api should ideally enforce this necessary ordering, or we should handle the case better when plugins are added after the bridge is attached to the webview.

PR #205, opened 2020-05-15
Added Bridge Builder classes, extension functions to clean up api

mattkranzler5:
The purpose of this PR is to clean up the API for creating/managing a `Bridge` instance. With the previous API it was possible to call `Bridge.add()` to add `Binders` after the bridge had already been attached which meant that those `Binders` would never get bound to the javascript engine. 

Here is an example of this: 
```
val bridge = WebViewBridge()
bridge.attach(webView)
bridge.add(SomePluginBinder(SomePlugin()))
```

The goal was to not allow this but also make the api a bit easier to read/use.

A `Builder` class was added to each `Bridge` implementation and the `Bridge.add()` and `Bridge.attach()` interface functions were removed and moved to the `Builder`. Now, this is what it looks like to create a bridge:

```
val bridge = WebViewBridge.Builder()
    .bind(SomePluginBinder(SomePlugin()))
    .attach(webView)
```

Another nice api change is that there is now an extension function generated on each `Plugin` to create a `Binder` for the plugin. This allows you to access the `Binder` like this:

```
// create WebViewBridge
val webViewBridge = WebViewBridge.Builder()
    .bind(SomePlugin().webViewBinder())
    .attach(webView)

// create V8Bridge
val v8Bridge = V8Bridge.Builder()
    .bind(SomePlugin().v8Binder())
    .attach(v8)
```

Finally, a couple extension functions were added to `WebView` and `V8` to allow creating a bridge in a more Kotlin DSL appropriate way:

```
// create WebViewBridge
val webViewBridge = webView.bridge {
    bind { SomePlugin().webViewBinder() }
}

// create V8Bridge
val v8Bridge = v8.bridge {
    bind { SomePlugin().v8Binder() }
}
```

Using the extension functions also allow the user to not specify `attach()` since the extension function automatically attaches to the receiver (`WebView` or `V8`).

### DIFF
--- a/platforms/android/bridge-v8/src/main/java/com/salesforce/nimbus/bridge/v8/V8Extensions.kt
+++ b/platforms/android/bridge-v8/src/main/java/com/salesforce/nimbus/bridge/v8/V8Extensions.kt
@@ -51,3 +51,11 @@ fun V8.rejectPromise(error: String): V8Object {
  * Helper function to create a [V8Object] which aids in testability.
  */
 fun V8.createObject() = V8Object(this)
+
+/**
+ * Creates a [V8Bridge.Builder] and passes it to the [builder] function, allowing any binders to be added
+ * and then attaches to the [V8Bridge] instance.
+ */
+fun V8.bridge(builder: V8Bridge.Builder.() -> Unit = {}): V8Bridge {
+    return V8Bridge.Builder().apply(builder).attach(this)
+}
--- a/platforms/android/bridge-v8/src/main/java/com/salesforce/nimbus/bridge/v8/V8Bridge.kt
+++ b/platforms/android/bridge-v8/src/main/java/com/salesforce/nimbus/bridge/v8/V8Bridge.kt
@@ -18,50 +18,12 @@ const val INTERNAL_NIMBUS_BRIDGE = "_nimbus"
 
 class V8Bridge : Bridge<V8, V8Object>, Runtime<V8, V8Object> {
 
-    var bridgeV8: V8? = null
-        private set
-    private val binders = mutableListOf<Binder<V8, V8Object>>()
+    private var bridgeV8: V8? = null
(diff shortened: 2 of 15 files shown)


threeve:
This should fix #111 

@paultiarks any thoughts about applying a similar pattern on iOS?

platforms/android/bridge-v8/src/main/java/com/salesforce/nimbus/bridge/v8/V8Extensions.kt
```diff
@@ -51,3 +51,11 @@ fun V8.rejectPromise(error: String): V8Object {
  * Helper function to create a [V8Object] which aids in testability.
  */
 fun V8.createObject() = V8Object(this)
+
+/**
+ * Creates a [V8Bridge.Builder] and passes it to the [builder] function, allowing any binders to be added
+ * and then attaches to the [V8Bridge] instance.
+ */
+fun V8.bridge(builder: V8Bridge.Builder.() -> Unit = {}): V8Bridge {
```
jamie-houston, line 59: What do you think about calling this `createBridge` or `attachBridge` (or something) so it's more obvious that this is the extension method?  Otherwise it's not immediately obvious (to me) when `bridge` is the package name or the extension.

mattkranzler5: The naming was intentional so that the extension function matched up with the naming of the `Bridge` class. If you feel strongly that it should be changed we can change it but I'd prefer to keep it as is since I think it makes the api look a bit nicer.

jamie-houston: Nope.  I'm good with it as is.

PR #208, opened 2020-05-20
iOS Shared Testing

paultiarks:
This adds the tests added on Android intended to be shared by both platforms to iOS. They are executing in both the webview and JSCore contexts.

One discrepancy is with a test that has a callback with 3 parameters. We don't currently support more than 2 parameters to a callback on iOS, with the intent being there would be a parameter for the result and a parameter for an error. @mattkranzler5 I think we may want to bring Android in line with iOS and find some way to limit Android to only 2 parameters as well. Do you have any thoughts on that?

### DIFF
--- a/platforms/apple/Sources/NimbusTests/SharedTestsWebView.swift
+++ b/platforms/apple/Sources/NimbusTests/SharedTestsWebView.swift
@@ -1,0 +1,117 @@
+//
+//  SharedTestsWebView.swift
+//  NimbusTests
+//
+//  Created by Paul Tiarks on 5/19/20.
+//  Copyright © 2020 Salesforce.com, inc. All rights reserved.
+//
+
+import XCTest
+import WebKit
+@testable import Nimbus
+
+class SharedTestsWebView: XCTestCase {
+    var webView = WKWebView()
+    var bridge = WebViewBridge()
+    var expectPlugin = ExpectPlugin()
+    var testPlugin = TestPlugin()
+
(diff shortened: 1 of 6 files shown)


platforms/apple/Sources/NimbusTests/SharedTestsWebView.swift
```diff
@@ -0,0 +1,116 @@
...
+        expectPlugin.reset()
+        expectPlugin.finishedExpectation = expectation(description: testName)
+        webView.evaluateJavaScript(testName, completionHandler: nil)
+        waitForExpectations(timeout: 1, handler: nil)
+        XCTAssertTrue(expectPlugin.isFinished)
+        XCTAssertTrue(expectPlugin.passed, "Failed: \(testName)")
+    }
+
+    func testAllTests() { //swiftlint:disable:this function_body_length
+        executeTest("verifyNullaryResolvingToInt()")
```
paultiarks, line 73: This function is curious. The reason for this is because binding a plugin with this many bound functions was causing the test to sometimes timeout setting up the webview (even with a 60 second timeout). I think this is because we're currently creating a WKUserScript per bound function. This could probably be reduced to a WKUserScript per plugin. I'm going to create a story to add this improvement and that should enable this test to look just like the JSCore tests.

mattkranzler5:
@paultiarks that makes sense (re: 2 parameters). I'll add a compiler check for that in Android and update the tests.

PR #209, opened 2020-05-21
Now limiting callbacks to two parameters, fixed bug in demo app

mattkranzler5:
This PR limits the amount of parameters in a function callback to 2. This brings Android in line with iOS in this respect. The intent is that one parameter is meant for the callback result and the other is the callback error.

### DIFF
--- a/platforms/android/compiler-v8/src/main/java/com/salesforce/nimbus/bridge/v8/compiler/V8BinderGenerator.kt
+++ b/platforms/android/compiler-v8/src/main/java/com/salesforce/nimbus/bridge/v8/compiler/V8BinderGenerator.kt
@@ -351,6 +351,12 @@ class V8BinderGenerator : BinderGenerator() {
         parameterIndex: Int
     ): CodeBlock {
 
+        // Check if there are more than two parameters in callback. Only two parameters (result, error) are allowed.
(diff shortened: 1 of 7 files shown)


platforms/android/compiler-v8/src/main/java/com/salesforce/nimbus/bridge/v8/compiler/V8BinderGenerator.kt
```diff
@@ -351,6 +351,12 @@ class V8BinderGenerator : BinderGenerator() {
         parameterIndex: Int
     ): CodeBlock {
 
+        // Check if there are more than two parameters in callback. Only two parameters (result, error) are allowed.
+        if (declaredType.typeArguments.size > 3) { // one type is for the return type (should be void)
```
jamie-houston, line 355: Do we check anywhere that the 3rd type is void? Should we add that check here as well?

mattkranzler5: There is a check for that already, on line 98 in `WebViewBinderGenerator` and line 195 in `V8BinderGenerator`.

PR #210, opened 2020-06-01
iOS functions with callbacks and a non-void return

afischsf:
Added the ability for all binding functions with callbacks to have a return value.

This is ONLY for 2.0.  Will be making another PR with the fixes for 1.0.

### DIFF
--- a/packages/test-www/test/shared-tests.js
+++ b/packages/test-www/test/shared-tests.js
@@ -498,24 +498,25 @@ function verifyBinaryIntDoubleResolvingToIntDoubleCallback() {
   }).then(() => {});
 }
 
-function verifyBinaryIntResolvingIntCallbackReturnsInt() {
-  let count = 0;
-  const verifyCallbacks = () => {
-    count = count + 1;
-    if (count === 2) {
-      __nimbus.plugins.expectPlugin.pass();
-      __nimbus.plugins.expectPlugin.finished();
-    }
-  }
-  __nimbus.plugins.testPlugin.binaryIntResolvingIntCallbackReturnsInt(3, (int) => {
-    if (int === 2) {
-      verifyCallbacks();
-    }
-  }).then((result) =>{
-    if (result === 1) {
-      verifyCallbacks();
-    }
-  });
-}
+// comment out unsupported test for android
+// function verifyBinaryIntResolvingIntCallbackReturnsInt() {
+//   let count = 0;
+//   const verifyCallbacks = () => {
+//     count = count + 1;
+//     if (count === 2) {
+//       __nimbus.plugins.expectPlugin.pass();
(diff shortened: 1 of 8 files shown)


packages/test-www/test/shared-tests.js
```diff
@@ -498,4 +498,24 @@ function verifyBinaryIntDoubleResolvingToIntDoubleCallback() {
   }).then(() => {});
 }
 
+function verifyBinaryIntResolingIntCallbackReturnsInt() {
+  let count = 0;
+  const test = () => {
```
paultiarks, line 503: Maybe a more descriptive name than `test`?

packages/test-www/test/shared-tests.js
```diff
@@ -498,4 +498,24 @@ function verifyBinaryIntDoubleResolvingToIntDoubleCallback() {
...
+    count = count + 1;
+    if (count === 2) {
+      __nimbus.plugins.expectPlugin.pass();
```
paultiarks, line 506: It really feels like `finished` should be called here since this test isn't done until both the return promise and the callback are called, or it times out.

packages/test-www/test/shared-tests.js
```diff
@@ -498,4 +498,24 @@ function verifyBinaryIntDoubleResolvingToIntDoubleCallback() {
   }).then(() => {});
 }
 
+function verifyBinaryIntResolingIntCallbackReturnsInt() {
```
paultiarks, line 501: nit: `Resoling` should be `Resolving`

platforms/apple/Sources/NimbusTests/SharedTestsJSCore.swift
```diff
@@ -212,4 +212,8 @@ class SharedTestsJSCore: XCTestCase {
     func testVerifyBinaryIntDoubleResolvingToIntDoubleCallback() {
         executeTest("verifyBinaryIntDoubleResolvingToIntDoubleCallback()")
     }
+    
+    func testVerifyBinaryIntResolingIntCallbackReturnsInt() {
```
paultiarks, line 216: This test should also be added to `SharedTestsWebView.swift`

packages/test-www/test/shared-tests.js
afischsf: RIP my speeeeling

packages/test-www/test/shared-tests.js
afischsf: I put pass here so the test would fail if both the callback function and the promise were not completed.  The idea being the test it self finishes when the promise is resolved, but it could fail if both functions were not touched.

Should the `finished` be right after `pass` in your opinion?  I am happy to change it to whatever you think is correct.

PR #217, opened 2020-06-11
Rename nimbus-bridge package to nimbus-types

threeve:


packages/test-www/test/index.ts
```diff
@@ -5,12 +5,12 @@
...
 //
 
-import "nimbus-bridge";
+import nimbus from "nimbus-types";
 import "./nimbus-core-tests";
 import "./broadcast-tests";
 import "./callback-encodable-tests";
 
-const { plugins } = __nimbus;
+const { plugins } = nimbus;
```
wes566, line 13: Discussed offline with @threeve... I have concerns around the name `nimbus-types` if the module has more than just types in it... for now this is fine though, there will be a future PR to split into 2 modules, one with the types and one with the implementation.

PR #224, opened 2020-06-15
Promise rejection fix

afischsf:
Rejected promises were sending the `undefined` data of a failed resolve type.
Now sending the `error` instead.

### DIFF
--- a/packages/nimbus-bridge/src/index.ts
+++ b/packages/nimbus-bridge/src/index.ts
@@ -136,13 +136,17 @@ let promisify = (src: any): void => {
   let dest: any = {};
   Object.keys(src).forEach((key): void => {
     let func = src[key];
-    dest[key] = (...args: any[]): any => {
+    dest[key] = (...args: any[]): Promise<any> => {
       args = cloneArguments(args);
-      let result = func.call(src, ...args);
-      if (result !== undefined) {
-        result = JSON.parse(result);
+      try {
+        let result = func.call(src, ...args);
+        if (result !== undefined) {
+          result = JSON.parse(result);
+        }
+        return Promise.resolve(result);
+      } catch (error) {
+        return Promise.reject(error);
       }
-      return Promise.resolve(result);
     };
   });
   return dest;
--- a/packages/nimbus-bridge/src/index.ts
+++ b/packages/nimbus-bridge/src/index.ts
@@ -162,7 +162,7 @@ let releaseCallback = (callbackId: string): void => {
 // in the storage
 let resolvePromise = (promiseUuid: string, data: any, error: any): void => {
   if (error) {
-    uuidsToPromises[promiseUuid].reject(data);
+    uuidsToPromises[promiseUuid].reject(error);
   } else {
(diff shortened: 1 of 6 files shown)


packages/nimbus-bridge/src/index.ts
```diff
@@ -162,7 +162,7 @@ let releaseCallback = (callbackId: string): void => {
 // in the storage
 let resolvePromise = (promiseUuid: string, data: any, error: any): void => {
   if (error) {
-    uuidsToPromises[promiseUuid].reject(data);
+    uuidsToPromises[promiseUuid].reject(error);
```
threeve, line 165: 👍  nice catch

platforms/apple/Sources/Nimbus/Connection.swift
```diff
@@ -141,7 +141,18 @@ public class Connection<C>: Binder {
...
     private func rejectPromise(promiseId: String, error: Error) {
-        webView?.evaluateJavaScript("__nimbus.resolvePromise('\(promiseId)', undefined, '\(error)');")
+        switch error {
+        case let encodableError as EncodableError:
+            let error = EncodableValue.error(encodableError)
+            if let data = try? JSONEncoder().encode(error),
+                let jsonString = String(data: data, encoding: .utf8) {
+                webView?.evaluateJavaScript("__nimbus.resolvePromise('\(promiseId)', undefined, \(jsonString).e);")
+            } else {
+                fatalError("Unable to encode error object: \(error)")
```
threeve, line 151: Should this `fatalError` or just fall through to the case where we send the error's description instead?

afischsf: Good question.  It would probably be best to send the description.  I will have to look into what a struct/class which is implementing Error & Encodable have for that property.

platforms/apple/Sources/Nimbus/Connection.swift
```diff
@@ -141,7 +141,19 @@ public class Connection<C>: Binder {
...
-        webView?.evaluateJavaScript("__nimbus.resolvePromise('\(promiseId)', undefined, '\(error)');")
+        switch error {
+        case let encodableError as EncodableError:
+            let error = EncodableValue.error(encodableError)
+            if let data = try? JSONEncoder().encode(error),
+                let jsonString = String(data: data, encoding: .utf8) {
+                webView?.evaluateJavaScript("__nimbus.resolvePromise('\(promiseId)', undefined, \(jsonString).e);")
+            } else {
+                //if unable to decode then send then fallthrough to default error message
+                fallthrough
```
afischsf, line 152: @threeve how about this?

threeve:
@afischsf what's the current status of this PR? is it failing tests on android because the expected errors do not match? Should we close this for now and roll this work into the changes that will come based on your error handling RFC?

threeve:
Maybe we can merge this by just having the tests verify that the promise correctly rejects, and we can check the error shape once we have the RFC implemented?

PR #225, opened 2020-06-24
Fix for passing a thrown error

afischsf:
When throwing an error in iOS it always returned a string of `undefined` because it was returning the data of a resolved promise instead of the error of a rejected one.


paultiarks:
Could we add a test that covers this?

PR #226, opened 2020-06-25
Plugin Event implementation for iOS

paultiarks:
This is the iOS implementation of the Plugin Events RFC. most of the changes are adding support for binding a callback which takes an `Encodable`. Includes additions to shared tests to verify the same behavior across platforms.

### DIFF
--- a/platforms/apple/Sources/Nimbus/EventPublisher.swift
+++ b/platforms/apple/Sources/Nimbus/EventPublisher.swift
@@ -17,19 +17,24 @@ private typealias ListenerMap = [String: Listener]
 
 class EventPublisher<Events: EventKeyPathing> {
     private var listeners: [String: ListenerMap] = [:]
+    private var listenerQueue: DispatchQueue = DispatchQueue(label: "EventPublisher")
 
     func addListener(name: String, listener: @escaping (Encodable) -> Void) -> String {
         let listenerId = UUID().uuidString
-        var listenerMap: ListenerMap = listeners[name, default: [:]]
-        listenerMap[listenerId] = listener
-        listeners[name] = listenerMap
+        listenerQueue.sync {
+            var listenerMap: ListenerMap = listeners[name, default: [:]]
+            listenerMap[listenerId] = listener
+            listeners[name] = listenerMap
+        }
         return listenerId
     }
 
     func removeListener(listenerId: String) {
-        listeners.forEach { key, map in
-            listeners[key] = map.filter { key, _ in
-                key != listenerId
+        listenerQueue.sync {
+            listeners.forEach { key, map in
+                listeners[key] = map.filter { key, _ in
+                    key != listenerId
+                }
             }
         }
     }
@@ -40,8 +45,10 @@ class EventPublisher<Events: EventKeyPathing> {
             return
         }
 
-        map.forEach { _, listener in
-            listener(payload)
+        listenerQueue.sync {
+            map.forEach { _, listener in
+                listener(payload)
+            }
         }
     }
 
--- a/platforms/apple/Sources/Nimbus/EventPublisher.swift
+++ b/platforms/apple/Sources/Nimbus/EventPublisher.swift
@@ -15,7 +15,7 @@ protocol EventKeyPathing {
 private typealias Listener = (Encodable) -> Void
 private typealias ListenerMap = [String: Listener]
 
-class EventTarget<Events: EventKeyPathing> {
+class EventPublisher<Events: EventKeyPathing> {
     private var listeners: [String: ListenerMap] = [:]
 
     func addEventListener(name: String, listener: @escaping (Encodable) -> Void) -> String {
--- a/platforms/apple/Sources/Nimbus/EventPublisher.swift
+++ b/platforms/apple/Sources/Nimbus/EventPublisher.swift
@@ -8,17 +8,34 @@
 
 import Foundation
 
-protocol EventKeyPathing {
+/**
+ Conformers to this protocol provide a mapping from a key path to a string
+ representation of the property name
+ */
+public protocol EventKeyPathing {
     static func stringForKeyPath(_ keyPath: PartialKeyPath<Self>) -> String?
 }
 
 private typealias Listener = (Encodable) -> Void
 private typealias ListenerMap = [String: Listener]
 
-class EventPublisher<Events: EventKeyPathing> {
+/**
+ A mix-in class to add event publishing capabilities to a plugin.
+
+ Instantiate with a type conforming to EventKeyPathing and call
+ this instance's bind function in your plugin's bind, passing in the
+ Connection, and javascript consumers can subscribe/unsubscribe
+ to events in your plugin. Publish events by passing a keyPath with
+ appropriate object as value to send events to listeners of that event.
+ */
+public class EventPublisher<Events: EventKeyPathing> {
     private var listeners: [String: ListenerMap] = [:]
     private var listenerQueue: DispatchQueue = DispatchQueue(label: "EventPublisher")
 
+    /**
+     Adds a listener for the given event name. This function is bound
+     in the bind method and shouldn't be called directly.
+     */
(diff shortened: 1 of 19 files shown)


platforms/apple/Sources/Nimbus/EventTarget.swift
```diff
@@ -0,0 +1,55 @@
...
+import Foundation
+
+protocol EventKeyPathing {
+    static func stringForKeyPath(_ keyPath: PartialKeyPath<Self>) -> String?
+}
+
+private typealias Listener = (Encodable) -> Void
+private typealias ListenerMap = [String: Listener]
+
+class EventTarget<Events: EventKeyPathing> {
```
threeve, line 18: this was renamed to `EventPublisher` in the TypeScript interface, we probably want to use the same nomenclature on native.

platforms/apple/Sources/Nimbus/WebViewConnection.swift
```diff
@@ -197,3 +210,9 @@ public class WebViewConnection: Connection, CallableBinder {
     private var bridge: JSEvaluating?
     private var bindings: [String: Callable] = [:]
 }
+
+extension Encodable {
+    func toJSONValue() throws -> Any? {
```
threeve, line 215: I'd prefer to inline this since it's only used in one location

platforms/apple/Sources/Nimbus/EventTarget.swift
```diff
@@ -0,0 +1,55 @@
...
+    private var listeners: [String: ListenerMap] = [:]
+
+    func addEventListener(name: String, listener: @escaping (Encodable) -> Void) -> String {
```
threeve, line 21:
```suggestion
    func addListener(name: String, listener: @escaping (Encodable) -> Void) -> String {
```

platforms/apple/Sources/Nimbus/EventTarget.swift
```diff
@@ -0,0 +1,55 @@
...
+        var listenerMap: ListenerMap = [:]
+        if let map = listeners[name] {
+            listenerMap = map
+        }
+        listenerMap[listenerId] = listener
+        listeners[name] = listenerMap
+        return listenerId
+    }
+
+    func removeEventListener(listenerId: String) {
```
threeve, line 32:
```suggestion
    func removeListener(listenerId: String) {
```

platforms/apple/Sources/Nimbus/EventTarget.swift
```diff
@@ -0,0 +1,55 @@
...
+            return
+        }
+
+        map.forEach { _, listener in
+            listener(payload)
+        }
+    }
+
+    func bind(to connection: Connection) {
+        connection.bind(addEventListener, as: "addEventListener")
```
threeve, line 52:
```suggestion
        connection.bind(addListener, as: "addListener")
```

platforms/apple/Sources/Nimbus/EventTarget.swift
```diff
@@ -0,0 +1,55 @@
...
+        connection.bind(removeEventListener, as: "removeEventListener")
```
threeve, line 53:
```suggestion
        connection.bind(removeListener, as: "removeListener")
```

platforms/apple/Sources/Nimbus/WebViewConnection.swift
```diff
@@ -123,6 +123,19 @@ public class WebViewConnection: Connection, CallableBinder {
         })
     }
 
+    func callbackEncodable(from value: Any?) -> Result<(Encodable) -> Void, Error> {
+        guard
+            let callbackId = value as? String,
+            let webView = self.webView else {
```
threeve, line 129:
```suggestion
            let webView = self.webView 
        else {
```

platforms/apple/Sources/Nimbus/WebViewConnection.swift
paultiarks: I don't think this can be in-lined, since it needs to be an extension on `Encodable` in order to compile?

threeve: I don't think I follow, you can't just put `try String(data: JSONEncoder().encode(value), encoding: .utf8)` in place of the call to this function?

paultiarks: Negative, that results in `Value of protocol type 'Encodable' cannot conform to 'Encodable'; only struct/enum/class types can conform to protocols`

threeve: Ahh bummer, I could have sworn that `encode` was not generic.

platforms/apple/Sources/Nimbus/EventPublisher.swift
```diff
@@ -0,0 +1,55 @@
...
+private typealias Listener = (Encodable) -> Void
+private typealias ListenerMap = [String: Listener]
+
+class EventPublisher<Events: EventKeyPathing> {
+    private var listeners: [String: ListenerMap] = [:]
+
+    func addListener(name: String, listener: @escaping (Encodable) -> Void) -> String {
+        let listenerId = UUID().uuidString
+        var listenerMap: ListenerMap = [:]
+        if let map = listeners[name] {
```
threeve, line 24: can probably replace this by using the subscript with default variant? `subscript(_:default:)`

```swift
var listenerMap = listeners[name, default: [:]]
```

platforms/apple/Sources/Nimbus/EventPublisher.swift
```diff
@@ -0,0 +1,55 @@
...
+            listenerMap = map
+        }
+        listenerMap[listenerId] = listener
+        listeners[name] = listenerMap
+        return listenerId
+    }
+
+    func removeListener(listenerId: String) {
+        listeners.forEach { key, map in
+            listeners[key] = map.filter { key, _ in
```
threeve, line 34: instead of iterating over every listener here could you instead do something like this?

```swift
listeners.forEach { key, map in map.removeValue(forKey: listenerId) }
```

platforms/apple/Sources/Nimbus/EventPublisher.swift
```diff
@@ -0,0 +1,55 @@
...
+        }
+    }
+
+    func publishEvent<V: Codable>(_ eventKeyPath: KeyPath<Events, V>, payload: V) {
+        guard let eventName = Events.stringForKeyPath(eventKeyPath),
+            let map = listeners[eventName] else {
+            return
+        }
+
+        map.forEach { _, listener in
```
threeve, line 46: do we need any sort of thread-safety around the listener map?

platforms/apple/Sources/Nimbus/EventPublisher.swift
paultiarks: `map` in that case is considered a let and is immutable.

threeve: how about something like this: 
```swift
listeners = listeners.mapValues { map in
    var updatedListeners = map
    updatedListeners.removeValue(forKey: listenerId)
    return updatedListeners
}```

I'm just trying to avoid the inner loops which isn't necessary because we know the key to remove.

File: platforms/apple/Sources/Nimbus/EventPublisher.swift [hunk 11]
```diff
@@ -0,0 +1,59 @@
...
+                listeners[key] = map.filter { key, _ in
+                    key != listenerId
+                }
+            }
+        }
+    }
+
+    func publishEvent<V: Codable>(_ eventKeyPath: KeyPath<Events, V>, payload: V) {
+        guard let eventName = Events.stringForKeyPath(eventKeyPath),
+            let map = listeners[eventName] else {
```
threeve, line 44: this access of `listeners` is not guarded by the `listenerQueue`, is that intentional?

paultiarks: nope, good catch

PR #228, opened 2020-06-29
Added an overload for the bind method.

sfdctaka:
Added one overload for the `bind` method. This overload is unique in that it takes a second callback. I am purposely omitting adding derivations of this overload because the use case of it is very limited for now. Should there be more use cases in the future that can take advantage of there being a second callback, then, overloading more `bind` methods should be revisited at that time. Android already supports the second callback via its `@BoundMethod` annotation.

### DIFF
--- a/platforms/apple/Sources/Nimbus/Binder.swift
+++ b/platforms/apple/Sources/Nimbus/Binder.swift
@@ -126,6 +126,14 @@ public protocol Binder {
     /**
      Bind the specified function to this connection.
      */
+    func bind<R: Encodable, A0, CB0: Encodable, CB1: Encodable>(
+        _ name: String,
+        to function: @escaping (A0, @escaping (CB0) -> Void, @escaping (CB1) -> Void) throws -> R
+    ) where A0: Decodable
+
+    /**
+     Bind the specified function to this connection.
+     */
     func bind<A0, A1, A2>(
         _ name: String,
         to function: @escaping (A0, A1, A2) throws -> Void
@@ -385,6 +393,16 @@ public extension Binder {
      */
     func bind<R: Encodable, A0, CB0: Encodable>(
         _ function: @escaping (A0, @escaping (CB0) -> Void) throws -> R,
+        as name: String
+    ) where A0: Decodable {
+        bind(name, to: function)
+    }
+
+    /**
+     Bind the specified function to this connection.
+     */
+    func bind<R: Encodable, A0, CB0: Encodable, CB1: Encodable>(
+        _ function: @escaping (A0, @escaping (CB0) -> Void, @escaping (CB1) -> Void) throws -> R,
         as name: String
     ) where A0: Decodable {
         bind(name, to: function)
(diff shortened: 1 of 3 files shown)


platforms/apple/Sources/NimbusTests/BinderTests.swift
```diff
@@ -591,6 +613,13 @@ class BindTarget {
         throw BindError.boundMethodThrew
     }
 
+    func ternaryWithTwoUnaryCallbackWithReturnThrows(arg0: Int, callback0: @escaping UnaryCallback, callback1: @escaping UnaryCallback) -> Int {
```
threeve, line 616: This test is named `Throws` but doesn't actually throw. Is it named wrong or is the body wrong?

platforms/apple/Sources/Nimbus/Binder.swift
```diff
@@ -123,6 +123,14 @@ public protocol Binder {
         to function: @escaping (A0, @escaping (CB0, CB1) -> Void) throws -> R
     ) where A0: Decodable
 
+    /**
+     Bind the specified function to this connection.
+     */
+    func bind<R: Encodable, A0, CB0: Encodable, CB1: Encodable>(
+        _ name: String,
+        to function: @escaping (A0, @escaping (CB0) -> Void, @escaping (CB1) -> Void) throws -> R
```
threeve, line 131: 👍 

This is good, but can you go ahead and see the 1-arg second callback added to the other places where it could accompany a 1-arg first callback, so basically the 2, 3, 4, and 5 arg `bind`s should each have the option for two 1-arg callbacks, and in both the void and return value cases. I'm ok leaving out the extra cases for a 2-arg second callback but I think our 1-arg overloads should be complete.

platforms/apple/Sources/Nimbus/CallableBinder.swift
```diff
@@ -218,6 +218,20 @@ extension CallableBinder {
...
+    public func bind<R: Encodable, A0, CB0: Encodable, CB1: Encodable>(
+        _ name: String,
+        to function: @escaping (A0, @escaping (CB0) -> Void, @escaping (CB1) -> Void) throws -> R
+    ) where A0: Decodable {
+        bindCallable(name) { [weak self] (args: [Any?]) in
+            guard let self = self else { throw DecodeError() }
+            try self.assertArgsCount(expected: 3, actual: args.count)
+            let a0 = try self.decode(args[0], as: A0.self).get()
+            let firstCallback = try self.callback(from: args[1], taking: CB0.self).get()
+            let seconCallback = try self.callback(from: args[2], taking: CB1.self).get()
```
threeve, line 230: There's a typo here for "second", but I would suggest renaming these to `callback0` and `callback1` or even `cb0` and `cb1` to mirror the type parameter names similar to how other arguments are named.

platforms/apple/Sources/NimbusTests/BinderTests.swift
sfdctaka: It was named wrong. Will take care!

platforms/apple/Sources/Nimbus/CallableBinder.swift
sfdctaka: 😬  On it.

sfdctaka:
I am closing this PR from my fork and file a new one from the origin to address the issue with CircleCI. I have forgotten about this. Will make all the changes in the new PR.

PR #230, opened 2020-07-01
Publishing Snapshots

jamie-houston:
- Updated publishing to allow snapshots.
- Moved nimbus modules into /modules
- Added buildsrc
- converted build.gradle to kts and removed redundant/unneeded config settings

Still left to do (in this PR or another one).  The first one is the only important one.  The others I don't see as urgent.
- reenable jacoco/coverage
- update deprecated (kotlintest) tests
- update deprecated kotlinpoet calls
- fix nimbusjs publish to not run if files aren't updated.
- add publishSnapshot task that appends SNAPSHOT to the version in  lerna.json.  Right now you have to update that manually and run artifactoryPublish
- readme on publishing and consuming
- cleanup remaining gradle files... still some logic split across gradle files (especially publish/artifactory).  This could be cleaned up and re-used across our repo/projects.


platforms/android/build.gradle
```diff
@@ -1,44 +1,36 @@
...
-
-    println('PREDEX ' + (project.ext.preDexLibs ? 'ENABLED' : 'DISABLED')) // goes to build log or console output
+plugins{
+    id("de.fayard.buildSrcVersions") version "0.7.0"
+//    id("com.autonomousapps.dependency-analysis") version "0.49.0"
+    id("com.jfrog.artifactory")
+    id("maven-publish")
+//    id("com.vanniktech.android.junit.jacoco") version "0.16.0"
 }
+//project.ext {
```
jamie-houston, line 25: this looks like it was a fix for a [circleci bug](https://stackoverflow.com/a/31877655/246914) that isn't an issue anymore so I'm removing it.  If we still need it I'll add it back.

platforms/android/gradle.properties
```diff
@@ -1,11 +1,10 @@
-android.enableJetifier=true
+android.enableJetifier=false
```
jamie-houston, line 1: I verified we can drop jetifier as no dependencies need it.  See https://github.com/plnice/can-i-drop-jetifier

platforms/android/gradle.properties
```diff
@@ -1,11 +1,10 @@
...
 android.useAndroidX=true
 org.gradle.jvmargs=-Xmx1536m
 
 # Publishing settings
 GROUP=com.salesforce.nimbus
-# This will get overridden by each modules build
-VERSION=0.0.1
+VERSION=2.0.12-SNAPSHOT
```
threeve, line 7: Does this mean the version now has to be set manually in this file every time and is no longer read from the top-level lerna json?

jamie-houston: No... just haven't changed this back yet.  Still cleaning this up.

platforms/android/modules/nimbusjs/src/main/res/raw/nimbus.js
```diff
@@ -0,0 +1,222 @@
+var __nimbus = (function () {
```
threeve, line 1: This file shouldn't be committed, it's a generated artifact from the build

platforms/android/shared-tests/build.gradle
```diff
@@ -1,20 +1,16 @@
-import groovy.json.JsonSlurper
-
-apply plugin: 'com.android.library'
-apply plugin: 'kotlin-android'
-apply plugin: 'kotlin-kapt'
-apply plugin: 'kotlinx-serialization'
+plugins {
```
mattkranzler5, line 1: Any reason why this file wasn't converted to `.kts`?

platforms/android/build.gradle
```diff
@@ -1,43 +1,27 @@
...
-        kotlinpoet_version = '1.5.0'
-        kotlin_test_runner_version = '3.1.5'
-        google_truth_version = '1.0.1'
-        j2v8_version = '6.1.0@aar'
-        androidx_test_version = '1.1.1'
-        espresso_version = '3.2.0'
-        mockk_version = '1.9.3.kotlin12'
-        json_version = '20190722'
-    }
+import groovy.json.JsonSlurper
```
mattkranzler5, line 1: Any reason why this file wasn't converted to `.kts`?

platforms/android/shared-tests/build.gradle
jamie-houston: The two that didn't get converted were because it wasn't a simple conversion and they didn't use any of the new buildSrc utils I think.  I can take a look and see if it's much work.  I'd like to convert all of the groovy files.  There are a few other gradle files to convert as well, like the publishing ones I added (semi) temporarily.

PR #231, opened 2020-07-01
WKUserScript Optimization

paultiarks:
This changes WebViewConnection to only generate a single WKUserScript per plugin. This also changes SharedTestsWebView to use a single test function per test case, rather than batching all test cases into a single function.

### DIFF
--- a/platforms/apple/Sources/Nimbus/WebViewConnection.swift
+++ b/platforms/apple/Sources/Nimbus/WebViewConnection.swift
@@ -58,21 +58,10 @@ public class WebViewConnection: Connection, CallableBinder {
      */
     func bindCallable(_ name: String, to callable: @escaping Callable) {
         bindings[name] = callable
-        let stubScript = """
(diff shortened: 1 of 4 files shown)


platforms/apple/Sources/Nimbus/WebViewConnection.swift
```diff
@@ -58,21 +58,10 @@ public class WebViewConnection: Connection, CallableBinder {
...
-          }
-          exports.push("\(name)");
-        }());
-        true;
-        """
+        scriptParts.append(name)
 
-        let script = WKUserScript(source: stubScript, injectionTime: .atDocumentStart, forMainFrameOnly: false)
-        webView?.configuration.userContentController.addUserScript(script)
+//        let script = WKUserScript(source: stubScript, injectionTime: .atDocumentStart, forMainFrameOnly: false)
```
threeve, line 63: remove?

PR #232, opened 2020-07-02
Android Event Publishing

mattkranzler5:
This is the Android event publishing implementation matching the Plugin Event RFC. There is a new interface called `EventPublisher` which should be implemented by a `Plugin` that wants to support event publishing. The plugin can implement the methods in the interface, however most implementations will just want to delegate to the `DefaultEventPublisher` which handles managing listeners and publishing events. This is what that property delegation looks like:
```
class MyPlugin : Plugin, EventPublisher<MyEvent> by DefaultEventPublisher() {...}
```
There is an interface called `Event` which each concrete event class needs to implement. Also, the implementation of `Event` must be `@Serializable` which is enforced by the compiler.

Here is an example of defining a single event class:
```
class MyEvent : Event {
    override val name: String = "myEvent"
}
```
The implementation must override the `name` property of `Event` to provide the name of the event for which it can be listened for when registering a listener, like this:
```
__nimbus.plugins.MyPlugin.addListener("myEvent", (event) => {...});
```
This implementation also supports multiple event types in a single publisher. This is what that might look like:
```
sealed class MyEvents : Event {

    @Serializable
    data class EventOne(val someString: String): MyEvents() {
        override val name: String = "eventOne"
    }

    @Serializable
    data class EventTwo(val someInt: Int): MyEvents() {
        override val name: String = "eventTwo"
    }
}
```

Then, a plugin can publish the event to listeners like this:
```
    val myPlugin = MyPlugin()
    myPlugin.publishEvent(EventOne())
    myPlugin.publishEvent(EventTwo())
```

### DIFF
--- a/platforms/android/compiler-base/src/main/java/com/salesforce/nimbus/compiler/BinderGenerator.kt
+++ b/platforms/android/compiler-base/src/main/java/com/salesforce/nimbus/compiler/BinderGenerator.kt
@@ -160,6 +160,17 @@ abstract class BinderGenerator : AbstractProcessor() {
         val pluginName = pluginElement.getAnnotation(PluginOptions::class.java).name
         val pluginTypeName = pluginElement.asKotlinTypeName()
 
+        // get event type if plugin is an event publisher
+        val eventType = types.directSupertypes((pluginElement.asType()))
+            .find { it.toString().startsWith("$nimbusPackage.EventPublisher") }?.typeArguments()?.firstOrNull()
+
+        if (eventType != null && !eventType.isKotlinSerializableType()) {
+            error(
+                pluginElement,
+                "$eventType is not @Serializable. Only Events that are @Serializable are currently supported."
+            )
+        }
+
         // read kotlin metadata so we can determine which types are nullable
         val kotlinClass =
             pluginElement.getAnnotation(Metadata::class.java)?.let { metadata ->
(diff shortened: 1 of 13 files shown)


platforms/android/compiler-base/src/main/java/com/salesforce/nimbus/compiler/BinderGenerator.kt
```diff
@@ -183,8 +194,16 @@ abstract class BinderGenerator : AbstractProcessor() {
...
+            .filter { it.kind == ElementKind.METHOD }
             .filter {
-                it.kind == ElementKind.METHOD && it.getAnnotation(BoundMethod::class.java) != null
+
+                // keep bound methods
+                it.getAnnotation(BoundMethod::class.java) != null ||
+
+                    // keep event publisher methods
+                    eventType != null &&
+                        (it.toString().startsWith("addListener") ||
```
threeve, line 205: Don't love this but we already discussed it and it is something we can hopefully change when we move to a kotlin compiler plugin rather than a java annotation processor.

platforms/android/core/src/test/java/com/salesforce/nimbus/DefaultEventPublisherTest.kt
```diff
@@ -0,0 +1,106 @@
...
+    private lateinit var eventOneListenerId: String
+    private var eventTwoListenerCalledWithCorrectData = false
+    private lateinit var eventTwoListenerId: String
+
+    @Before
+    fun setUp() {
+        eventPublisher = DefaultEventPublisher<TestEvents>().apply {
+            eventOneListenerId = addListener("testEventOne") {
+                eventOneListenerCalledWithCorrectData =
+                    it is TestEvents.TestEventOne && it.message == "testMessage"
```
jamie-houston, line 34: nit: use a UUID or random string for this instead of hard coded.

mattkranzler5: @jamie-houston I'm not sure what you are referring to. What do you want to use the UUID for?

jamie-houston: "testMessage"
```
val testMessage = UUID.randomUUID().toString()
...
    it is TestEvents.TestEventOne && it.message == testMessage
...
    eventPublisher.publishEvent(TestEvents.TestEventOne(message = testMessage))

```

mattkranzler5: That won't really work in this case without some major changes. The test message is created in the test and the listener is created in `setUp()`. The test is just to verify that the listener is invoked (the message isn't super relevant).

PR #233, opened 2020-07-02
Promise rejection fix (#224)

threeve:
* send the error when rejecting the promise

* iOS now correctly passes Error's to the promise rejection.
iOS rejection errors can now also be Encodable Error structures/classes.

* make the linter happy

* send interpolated error if unable to decode

* Ensure promises get rejected on android when an exception is thrown

Co-authored-by: Jason Foreman <jforeman@salesforce.com>

### DIFF
--- a/platforms/apple/Sources/Nimbus/JSContextConnection.swift
+++ b/platforms/apple/Sources/Nimbus/JSContextConnection.swift
@@ -52,7 +52,20 @@ public class JSContextConnection: Connection, CallableBinder {
                 }
                 return self.promiseGlobal?.invokeMethod("resolve", withArguments: resultArguments)
             } catch {
-                return self.promiseGlobal?.invokeMethod("reject", withArguments: [error.localizedDescription])
+                switch error {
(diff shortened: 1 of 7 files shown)


platforms/apple/Sources/Nimbus/JSContextConnection.swift
```diff
@@ -52,6 +52,19 @@ public class JSContextConnection: Connection, CallableBinder {
...
+                        let promiseGlobal = self.promiseGlobal,
+                        let jsValue = try? JSValueEncoder().encode(error, context: promiseGlobal.context),
+                        let errorValue = jsValue.objectForKeyedSubscript("e")
+                    {
+                        return promiseGlobal.invokeMethod("reject", withArguments: [errorValue])
+                    }
+                default:
+                    break
+                }
                 return self.promiseGlobal?.invokeMethod("reject", withArguments: [error.localizedDescription])
```
paultiarks, line 68: Maybe worth moving this up into the `default` block to match the WebViewConnection implementation?

PR #234, opened 2020-07-02
Bundle nimbus.js as an asset rather than raw resource (#182)

threeve:
* Bundle nimbus.js as an asset rather than raw resource


threeve:
This was changed on the `release-1.x` branch but hadn't yet been pulled into `main`

PR #235, opened 2020-07-02
iOS CI Update

paultiarks:
This runs tests on both iOS 12 & iOS 13 and adds a missing test case from the EventPublisher

### DIFF
--- a/platforms/apple/Sources/NimbusTests/CallJavascriptTests.swift
+++ b/platforms/apple/Sources/NimbusTests/CallJavascriptTests.swift
@@ -38,7 +38,7 @@ class CallJavascriptTests: XCTestCase, WKNavigationDelegate {
     func loadWebViewAndWait(html: String = "<html><body></body></html>") {
         loadingExpectation = expectation(description: "web view loaded")
         webView.loadHTMLString(html, baseURL: nil)
-        wait(for: [loadingExpectation!], timeout: 10)
+        wait(for: [loadingExpectation!], timeout: 60)
     }
 
     func webView(_: WKWebView, didFinish _: WKNavigation!) {
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,7 +60,62 @@ jobs:
           name: Build Swift Package
           command: swift build
 
-  build-and-test-ios:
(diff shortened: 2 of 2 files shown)


platforms/apple/Sources/NimbusTests/CallJavascriptTests.swift
```diff
@@ -38,7 +38,7 @@ class CallJavascriptTests: XCTestCase, WKNavigationDelegate {
     func loadWebViewAndWait(html: String = "<html><body></body></html>") {
         loadingExpectation = expectation(description: "web view loaded")
         webView.loadHTMLString(html, baseURL: nil)
-        wait(for: [loadingExpectation!], timeout: 10)
+        wait(for: [loadingExpectation!], timeout: 60)
```
threeve, line 41: 😩

platforms/apple/Sources/NimbusTests/EventPublisherTests.swift
```diff
@@ -64,6 +64,23 @@ class EventPublisherTests: XCTestCase {
         waitForExpectations(timeout: 1, handler: nil)
     }
 
+    func testDispatchTwice() {
```
threeve, line 67: Could you split this out into its own PR, it feels very out of place with what the focus of this PR is.

platforms/apple/Sources/NimbusTests/CallJavascriptTests.swift
paultiarks: It's truly a tragedy, but I don't know of an easier/cleaner way of solving the problem without introducing a bunch more risk.

threeve:
Do we need to test on iOS 11 as well, since that's our current deployment target?

threeve:
Or should we consider bumping deployment target to 12?

paultiarks:
I'm not sure dropping iOS 11 is necessary at this point. We should probably be testing all three. The build and test task is already fairly large and takes a while. I think I'm going to create two new tasks, 1 that builds and tests on 11, and 1 that builds and tests on 12. Then I'm going to amend some of the changes already done here and leave the main job doing all of the coverage gathering/uploading.

PR #243, opened 2020-07-14
Implementation of return value errors on iOS

paultiarks:
Most of this was already implemented in `main` but I found a problem in `JSValueEncoder` that was causing the encoding of `EncodableValue` to encode no values. This fixes that, fixes the implementation in JSCore for the simple, non-structured enum, and adds tests to the shared test library for both cases.

### DIFF
--- a/platforms/apple/Sources/Nimbus/JSValueEncoder.swift
+++ b/platforms/apple/Sources/Nimbus/JSValueEncoder.swift
@@ -71,9 +71,17 @@ class JSValueEncoderContainer: Encoder {
     var storage: JSValueEncoderStorage
     let context: JSContext
 
-    init(context: JSContext) {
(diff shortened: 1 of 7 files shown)


platforms/apple/Sources/Nimbus/JSValueEncoder.swift
```diff
@@ -71,11 +71,19 @@ class JSValueEncoderContainer: Encoder {
     var storage: JSValueEncoderStorage
     let context: JSContext
 
-    init(context: JSContext) {
+    init(userInfo: [CodingUserInfoKey: Any] = [:], codingPath: [CodingKey] = [], context: JSContext) {
```
sfdctaka, line 74: Do we need `userInfo` and `codingPath` for this initializer?

paultiarks: Yes, they were added for the `ReferencingEncoder` to instantiate a new encoder with the right context.

PR #244, opened 2020-07-15
Generate core plugins in consuming app with new annotation

mattkranzler5:
This PR creates a new annotation `@BoundPlugin` which allows apps to annotate library plugins (`DeviceInfoPlugin`, for example) when they are used as fields or in value parameters which will trigger binder generation for them. Previously the binders were being generated in the library which would cause transitive dependencies to be included in the consuming app. 

The consumer can declare a plugin as a class variable, like this:
```
class SomeActivity : Activity {
    @BoundPlugin 
    val plugin: SomeLibraryPlugin = SomeLibraryPlugin()
}
```
This will generate whatever binders the library supports as long as the compiler for that binder is included in the consuming app. For example, if the plugin supports both webview and v8 but the app does not include the `compiler-v8` dependency then no v8 binder will be generated. However, if the plugin does not support a bridge and the consuming app includes the compiler dependency for that binder it will also not be generated.

### DIFF
--- a/platforms/android/modules/core-plugins/build.gradle.kts
+++ b/platforms/android/modules/core-plugins/build.gradle.kts
@@ -11,16 +11,10 @@ android {
 }
 
 dependencies {
-    compileOnly(nimbusModule("core"))
-    compileOnly(nimbusModule("annotations"))
-    compileOnly(nimbusModule("bridge-webview"))
-    compileOnly(nimbusModule("bridge-v8"))
-    api(Libs.kotlinStdlib)
+    implementation(nimbusModule("core"))
+    implementation(nimbusModule("annotations"))
+    implementation(Libs.kotlinStdlib)
     compileOnly(Libs.kotlinxSerializationRuntime)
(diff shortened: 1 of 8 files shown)


platforms/android/modules/core-plugins/build.gradle.kts
```diff
@@ -11,16 +11,10 @@ android {
...
-    compileOnly(nimbusModule("bridge-v8"))
-    api(Libs.kotlinStdlib)
+    implementation(nimbusModule("core"))
+    implementation(nimbusModule("annotations"))
+    implementation(Libs.kotlinStdlib)
     compileOnly(Libs.kotlinxSerializationRuntime)
-    compileOnly(Libs.j2v8)
-    compileOnly(Libs.k2v8)
-    kapt(nimbusModule("compiler-webview"))
-    kapt(nimbusModule("compiler-v8"))
```
jamie-houston, line 23: Much better!

PR #245, opened 2020-07-16
CI for Pre-release builds

paultiarks:
This adds two new tasks to the circle ci config, both only triggered by a tag push where the tag matches something that looks like `major.minor.patch-beta(beta number)`.

One for iOS which uploads a release to GitHub for the given tag, marked as pre-release.

One that calls a new Android gradle task that massages the project version to append `-SNAPSHOT` and cut out any beta strings and then calls the `artifactoryPublish` task to upload a snapshot.

### DIFF
--- a/platforms/android/build.gradle.kts
+++ b/platforms/android/build.gradle.kts
@@ -70,7 +70,7 @@ tasks.register("publishSnapshot") {
     }
     doLast {
         val stringVersion = version.toString()
-        val regex = "[0-9]+.[0-9]+.[0-9]+".toRegex()
+        val regex = "[0-9]+\\.[0-9]+\\.[0-9]+".toRegex()
         val numericVersion = regex.find(stringVersion)
         if (numericVersion != null) {
             version = numericVersion.value + "-SNAPSHOT"
--- a/platforms/android/build.gradle.kts
+++ b/platforms/android/build.gradle.kts
@@ -1,6 +1,7 @@
 import groovy.json.JsonSlurper
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import org.jetbrains.dokka.gradle.DokkaTask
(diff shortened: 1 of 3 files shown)


platforms/android/build.gradle.kts
```diff
@@ -60,6 +61,26 @@ tasks {
...
+tasks.register("publishSnapshot") {
+    val publishTask = tasks.findByPath("artifactoryPublish")
+    if (publishTask != null) {
+        this.finalizedBy(publishTask)
+    } else {
+        throw GradleException("Unable to find the publish task")
+    }
+    doLast {
+        val stringVersion = version.toString()
+        val regex = "[0-9]+.[0-9]+.[0-9]+".toRegex()
```
jamie-houston, line 73: This will ignore the number in `beta[number]` correct?  Do we care about that or are ok with it just creating another snapshot with the same x.y.z version?

paultiarks: Yeah my intent here was to pick out the `x.y.z` and then just append SNAPSHOT to that. I don't think I have a strong preference, but I was under the impression we wanted to ignore the `beta` part.

threeve, line 73: Does the `.` not need to be escaped here?

paultiarks: Doesn't seem to be. It was working as expected when I tested it. And it's a direct copy of the regex we use in the Xcode build phase for setting the project version.

paultiarks: Actually, yeah, good call, I think it should be. Going to add it here *and* to the Xcode build phase

paultiarks: Updated

jamie-houston: That's correct.  For android snapshots at least, we don't care about the beta part.

PR #249, opened 2020-07-17
Android promise error handling

mattkranzler5:
This PR handles properly rejecting promises in Android with exceptions that occur on the native side, adding support for `@Serializable` exceptions. For normal non-declared exceptions the promise will be resolved with the `Exception.message`. A new property has been added to `@BoundMethod` called `throwsExceptions` which allows specifying one or more exceptions that the bound method can throw. If those exceptions are `@Serializable` the promise will be resolved with the serialized exception. 

Changes were necessary to the way synchronous returns are handled in the `WebView` binders in order to allow for handling/serializing the exception. `WebView` binders no longer will return a value and instead will invoke the `__nimbus.resolvePromise()` function with either the return value or the exception.

### DIFF
--- a/platforms/android/demo-app/src/main/java/com/salesforce/nimbusdemoapp/MainActivity.kt
+++ b/platforms/android/demo-app/src/main/java/com/salesforce/nimbusdemoapp/MainActivity.kt
@@ -29,6 +29,7 @@ import com.salesforce.nimbus.bridge.webview.WebViewBridge
 import com.salesforce.nimbus.bridge.webview.bridge
 import com.salesforce.nimbus.core.plugins.DeviceInfoPlugin
 import kotlinx.serialization.Serializable
+import java.lang.RuntimeException
 
 class MainActivity : AppCompatActivity() {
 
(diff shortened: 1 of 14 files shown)


platforms/android/demo-app/src/main/java/com/salesforce/nimbusdemoapp/MainActivity.kt
```diff
@@ -146,3 +174,32 @@ sealed class MessageEvents : Event {
...
+    @BoundMethod
+    fun throwException(throwException: Boolean): String {
+        return if (throwException) {
+            throw Exception("This is the exception")
+        } else {
+            "This is the return value"
+        }
+    }
+
+    @BoundMethod(SerializableException1::class, SerializableException2::class)
```
threeve, line 197: Would it be better to require this to be more explicit, eg
```kotlin
@BoundMethod(throws: [E1::class, E2::class])
```

From this side of things when reading the code it's not clear what the annotation parameters represent

mattkranzler5: Naming the parameters is optional. You can either do the above or declare them like this:
```
@BoundMethod(throwsExceptions = [E1::class, E2::class])
```
It's up to the user how they want to define them.

PR #250, opened 2020-07-20
[draft] fix(coverage): Add code coverage for all android modules with tests

jamie-houston:
- Remove code cov badge on readme (it was pointing to an old branch)
- codecov paths updated for all modules with tests
- gradle build fails in console with failed tests
- show output in console for pass/fail for tests
- PR should only fail if coverage drops from destination branch, not if it falls below 75%


jamie-houston:
Making changes.

PR #254, opened 2020-07-22
fix(publish) - set snapshot version when publishing snapshot

jamie-houston:
Fixes a bug in the publishSnapshot task that didn't set the version at the correct time in the build lifecycle, so version was staying as x.x.x-beta instead of x.x.x-SNAPSHOT

### DIFF
--- a/packages/nimbus-bridge/package-lock.json
+++ b/packages/nimbus-bridge/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "nimbus-types",
-    "version": "2.0.0",
+    "version": "2.0.0-beta1",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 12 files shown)


packages/nimbus-bridge/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "nimbus-types",
-    "version": "2.0.0",
+    "version": "2.0.0-beta1",
```
jamie-houston, line 3: should this match lerna.json version...?

platforms/android/modules/core-plugins/build.gradle.kts
```diff
@@ -11,6 +11,9 @@ plugins {
     kotlin("plugin.serialization")
     kotlin("kapt")
     id("org.jetbrains.dokka")
+    `maven-publish`
```
jamie-houston, line 14: I'm assuming we want to publish core-plugins...?  We weren't before.

packages/nimbus-bridge/package-lock.json
paultiarks: It ideally should. I think it didn't before because it doesn't update until the npm build runs, which I didn't run after updating the version.

PR #259, opened 2020-08-05
Ensure plugins are added before attaching the bridge to the webview

HighSide:
Create BridgeBuilder in order to control initialization order.


platforms/apple/Sources/Nimbus/BridgeBuilder.swift
```diff
@@ -0,0 +1,66 @@
...
+        context.evaluateScript(nimbusDeclaration)
+        
+        for plugin in plugins {
+            let connection = JSContextConnection(from: context, bridge: bridge, as: plugin.namespace)
+            plugin.bind(to: connection)
+        }
+    }
+}
+
+public class Bridge: NSObject {
```
paultiarks, line 64: Can this be a protocol instead of a base class?

platforms/apple/Sources/Nimbus/BridgeBuilder.swift
```diff
@@ -0,0 +1,66 @@
...
+            attach(bridge: bridge, webView: webView, plugins: plugins)
+            return bridge
+        case .jsContext(let context):
+            let bridge = JSContextBridge(context: context)
+            attach(bridge: bridge, context: context, plugins: plugins)
+            return Bridge()
+        }
+    }
+    
+    static func attach(bridge: WebViewBridge, webView: WKWebView, plugins: [Plugin]) {
```
paultiarks, line 32: Something to think about for this and the other `attach`, the implementation is very specific to each respective bridge type. What are the pros and cons of moving these into internal implementations on each bridge class?

platforms/apple/Sources/Nimbus/BridgeBuilder.swift
HighSide: I'll look into that...

PR #260, opened 2020-08-06
Fix an issue where nimbusjs was not executing the copyScript task

threeve:
This fixes @W-7924825@

### DIFF
--- a/platforms/android/modules/nimbusjs/build.gradle.kts
+++ b/platforms/android/modules/nimbusjs/build.gradle.kts
@@ -48,7 +48,7 @@ val npmInstallTask = tasks.named<NpmTask>("npm_install") {
 }
 
 tasks.whenTaskAdded {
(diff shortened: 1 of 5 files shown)


platforms/android/modules/nimbusjs/build.gradle.kts
```diff
@@ -48,7 +48,7 @@ val npmInstallTask = tasks.named<NpmTask>("npm_install") {
 }
 
 tasks.whenTaskAdded {
-    if (name.startsWith("assemble")) {
+    if (name.startsWith("generate") && name.endsWith("Assets")) {
```
threeve, line 51: This should attach to `generateDebugAssets` and `generateReleaseAssets` tasks, which is early enough in the build cycle that copying nimbus.js into the assets folder is picked up by later tasks

PR #262, opened 2020-08-10
Reorganize packages according to RFC.

threeve:
- Split nimbus-bridge into @nimbus-js/{api,runtime} packages


docs/package.json
```diff
@@ -14,10 +14,5 @@
     "type": "git",
     "url": "https://github.com/salesforce/nimbus"
   },
-  "license": "CC0-1.0",
-  "devDependencies": {
```
wes566, line 18: Can we keep each package's deps listed in the respective package.json file?  It's more accurate and explicit.  In fact, [lerna docs even call it a "problem" when a package.json doesn't explicitly list its deps](https://github.com/lerna/lerna/blob/master/doc/hoist.md#forgetting-dependencies).

packages/@nimbus-js/api/src/index.ts
```diff
@@ -0,0 +1,72 @@
+//
+// Copyright (c) 2019, Salesforce.com, inc.
```
wes566, line 2:
```suggestion
// Copyright (c) 2020, Salesforce.com, inc.
```

packages/@nimbus-js/runtime/src/nimbus.ts
```diff
@@ -285,15 +218,14 @@ window.addEventListener("unload", (): void => {
...
 
-declare global {
-  var __nimbus: Nimbus;
-}
-
 window.__nimbus = nimbus;
 
 export default nimbus;
+
+// TODO:
```
wes566, line 230: did you mean to leave this TODO in there?

packages/test-www/package.json
```diff
@@ -11,19 +12,6 @@
...
-        "chai": "^4.2.0",
-        "cross-env": "^5.2.0",
-        "fs-extra": "^7.0.1",
-        "mocha": "^6.1.4",
-        "rollup": "^1.16.6",
-        "rollup-plugin-node-resolve": "^5.2.0",
-        "rollup-plugin-serve": "^0.6.0",
-        "rollup-plugin-typescript2": "^0.21.2",
-        "typescript": "^3.3.3"
+        "@nimbus-js/api": "file:../@nimbus-js/api"
```
wes566, line 15: With lerna you shouldn't have to do this... you should be able to just use the version number and then `lerna bootstrap` will take care of linking everything

packages/test-www/rollup.config.js
```diff
@@ -21,7 +21,7 @@ fs.removeSync(dist);
 // Copy public folder
 fs.copySync(publicFolder, dist);
 
-const mochaPath = path.resolve(__dirname, "node_modules", "mocha");
+const mochaPath = path.resolve(__dirname, "..", "..", "node_modules", "mocha");
```
wes566, line 24:
```suggestion
const mochaPath = require.resolve("mocha");
```

packages/test-www/rollup.config.js
```diff
@@ -30,7 +30,7 @@ fs.copySync(
   path.resolve(mochaPath, "mocha.js"),
   path.resolve(dist, "mocha.js")
 );
-const chaiPath = path.resolve(__dirname, "node_modules", "chai");
+const chaiPath = path.resolve(__dirname, "..", "..", "node_modules", "chai");
```
wes566, line 33:
```suggestion
const chaiPath = require.resolve("chai");
```

packages/test-www/test/nimbus-core-tests.ts
```diff
@@ -111,32 +111,31 @@ describe("Nimbus JS API", () => {
...
           intField: 42,
-          stringField: "JSAPITEST",
+          stringField: "JSAPITEST"
         });
         done();
       }
     );
   });
 
   // commented out test because they fail in android as it is not supported yet
```
wes566, line 127: is this comment true still?  Looks like the test is no longer commented out

docs/package.json
threeve: Is this really a problem for `devDependencies` as well? I could potentially see moving stuff back like 11ty that is _only_ used for docs, but it's really handy having only one place to specify the TypeScript version and stuff like eslint, prettier, etc.

packages/@nimbus-js/runtime/src/nimbus.ts
threeve: Yeah this is intentional, since we're no longer going to have a build step this will be basically a manual IIFE.

packages/test-www/package.json
threeve: Lerna did this itself as the result of running `lerna link convert`.

packages/test-www/test/nimbus-core-tests.ts
threeve, line 127:
```suggestion
```

docs/package.json
wes566: Yea, for `devDependencies` as well... because the idea is a package/sub-project could be taken out of the monorepo and it still work (because its package.json correctly lists all deps).

In regards to your concern, [`lerna add`](https://github.com/lerna/lerna/tree/master/commands/add#examples) is supposed to help with this - `lerna add typescript@3.9.7` will add/update TS to v3.9.7 for all packages in the monorepo (of course you can filter which packages to add/update as well)

packages/test-www/package.json
wes566: I see, I just read the docs and that's what the `convert` arg does... I'm not sure you need to do that, but if that's the preference to just use file-based paths then that's fine too, I have no strong opinion about it.

threeve: So, I ran `lerna link convert` but I'm not sure that's what I wanted, what I was trying to do was "hoist" but link convert is different. I can change it back if you think it's better the other way?

wes566: My preference would be to go back to using version numbers and let `lerna bootstrap` do the linking magic.

PR #264, opened 2020-08-12
Fix null Metadata for the kotlin classes.

gatorboy:
The plugin classes' metadata was not retrieved with getAnnotation call. Because of this, it was unable to determine the nullability of the fields.

This fixes the metadata and generates nullable fields on callbacks and other function calls.


gatorboy:
@jamie-houston I've added tests. Could you review?

PR #265, opened 2020-08-13
Convert nimbus.ts to nimbus.js with jsdoc type annotations

threeve:
This will serve as a basis for @paultiarks to make the native side changes necessary to include nimbus.js in the core framework.


packages/@nimbus-js/runtime/package.json
```diff
@@ -4,18 +4,14 @@
...
-    "types": "dist/es/nimbus.d.ts",
+    "main": "src/nimbus.js",
     "files": [
-        "/dist"
+        "/src/nimbus.js"
     ],
     "scripts": {
-        "build": "rollup -c rollup.config.js",
-        "build-docs": "typedoc --out dist/typedoc --tsconfig ./tsconfig.json ./src",
+        "build": "tsc",
```
wes566, line 12: nit: if this is just type-checking and not actually building anything then maybe call it "check" or "typecheck"???

packages/@nimbus-js/runtime/src/nimbus.js
```diff
@@ -0,0 +1,276 @@
...
+      } else {
+        clonedArgs.push(args[i]);
+      }
+    }
+    return clonedArgs;
+  };
+
+  /**
+   * @param {*} src
+   * @returns {Object}
```
wes566, line 78: I feel like this return type could be more specific... 🤔 

I think the type you specify in line 81 could go into a typedef and then shared between line 81 and line 78.

packages/@nimbus-js/runtime/src/nimbus.js
```diff
@@ -0,0 +1,276 @@
...
+        }
+      }
+      if (found) {
+        messageListeners.splice(counter, 1);
+        eventNameToListeners[message] = messageListeners;
+      }
+    }
+  };
+
+  // { import("global.d.ts") }
```
wes566, line 193: can you remove this comment?

packages/@nimbus-js/runtime/src/nimbus.js
```diff
@@ -0,0 +1,276 @@
...
+    }
+    return clonedArgs;
+  };
+
+  /**
+   * @param {*} src
+   * @returns {Object}
+   */
+  let promisify = src => {
+    /** @type { {[s: string]: function(...any[]): void} } */
```
wes566, line 81: This type should have Promise somewhere in it right?

```suggestion
    /** @type { {[s: string]: function(...any[]): Promise<any>} } */
```

packages/@nimbus-js/runtime/src/nimbus.js
```diff
@@ -0,0 +1,276 @@
...
+      return (
+        asNumber ^
+        (crypto.getRandomValues(new Uint8Array(1))[0] & (15 >> (asNumber / 4)))
+      ).toString(16);
+    });
+  };
+
+  /**
+   * @param {any[]} args
+   * @returns {any[]}
```
wes566, line 58: nit: I've noticed this file uses `any` in some spots and `*` in some spots... maybe we normalize on using `any`?

packages/@nimbus-js/runtime/src/nimbus.js
```diff
@@ -0,0 +1,276 @@
...
+        [pluginName]: plugin
+      });
+    });
+  }
+  let nimbusBuilder = {
+    plugins: plugins
+  };
+
+  /** @type { import("@nimbus-js/api").Nimbus } */
+  let nimbus = Object.defineProperties(nimbusBuilder, {
```
wes566, line 243: Is there a reason this is `let` and not `const`?  I don't see where `nimbus` gets mutated anywhere.  Same goes for all the other variables in this file... I think most of them could be const right?  Maybe I'm missing something.

packages/@nimbus-js/runtime/tsconfig.json
```diff
@@ -1,4 +1,10 @@
 {
     "extends": "../../../tsconfig.json",
-    "include": ["src/**/*"]
+    "include": ["src/**/*"],
+
+    "compilerOptions": {
+        "noEmit": true,
+        "allowJs": true,
+        "typeRoots": ["./src/", "./node_modules/@types", "./node_modules/**"]
```
wes566, line 8: Do you need the `node_modules/**` here?

```suggestion
        "typeRoots": ["./src/", "./node_modules/@types"]
```

packages/@nimbus-js/runtime/package.json
threeve: I originally did have it called that, but I ended up changing it to `build` so that it would run when doing a build from the root of the lerna project.

packages/@nimbus-js/runtime/src/nimbus.js
threeve: I dunno the difference, this is just whatever the "document this" extension decided to put in for us ¯\_(ツ)_/¯

PR #266, opened 2020-08-13
fix: Generated binder extension methods now match visibility of the class being extended

nadeem-sfdc:
When generating a binder for an `internal` class, the nimbus compiler generates a 'public' extension method for creating the binder. This extension method takes an instance of the plugin class as an argument, which results in a compiler failure due to exposing an internal class in a public extension method. 

This fixes the issue and now the compiler generates the binder extension method with the same visibility as the plugin class it is extending.

### DIFF
--- a/platforms/android/modules/compiler-base/src/main/java/com/salesforce/nimbus/compiler/BinderGenerator.kt
+++ b/platforms/android/modules/compiler-base/src/main/java/com/salesforce/nimbus/compiler/BinderGenerator.kt
@@ -19,6 +19,9 @@ import com.squareup.kotlinpoet.PropertySpec
 import com.squareup.kotlinpoet.TypeName
 import com.squareup.kotlinpoet.TypeSpec
 import com.squareup.kotlinpoet.asClassName
+import kotlinx.metadata.Flag
+import kotlinx.metadata.Flags
+import kotlinx.metadata.KmClass
 import kotlinx.metadata.KmFunction
 import kotlinx.metadata.jvm.KotlinClassHeader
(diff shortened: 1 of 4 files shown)


platforms/android/modules/compiler-base/src/main/java/com/salesforce/nimbus/compiler/BinderGenerator.kt
```diff
@@ -220,14 +225,16 @@ abstract class BinderGenerator : AbstractProcessor() {
             }
             .map { it as ExecutableElement }
 
+        kotlinClass?.let { classModifier = processClassModifierTypes(it) }
```
jamie-houston, line 228: nit: to avoid a mutable variable, could this be something like:
```
val classModifier = kotlinClass?.let(::processClassModifierTypes) ?: KModifier.PUBLIC
```

platforms/android/modules/compiler-webview/src/main/java/com/salesforce/nimbus/bridge/webview/compiler/WebViewBinderGenerator.kt
```diff
@@ -80,9 +82,15 @@ class WebViewBinderGenerator : BinderGenerator() {
         // return type is nullable
         val kotlinReturnType = kotlinFunction?.returnType
 
+        // Set default Function modifier to Public
+        var funModifiers = KModifier.PUBLIC
+        kotlinFunction?.let { funModifiers = processFunctionModifierTypes(it) }
```
jamie-houston, line 87: same comment.  avoid mutable types if possible.

PR #267, opened 2020-08-14
NimbusJS removal and addition of nimbus.js to Nimbus

paultiarks:
As the description says, this adds nimbus.js as a resource to the main Nimbus target/module and removes NimbusJS podspec for iOS and nimbusjs module for android.


platforms/android/modules/core/build.gradle.kts
```diff
@@ -27,10 +27,23 @@ dependencies {
...
     testImplementation(Libs.kotlinxSerializationRuntime)
+    testImplementation(Libs.truth)
+    testImplementation("org.robolectric:robolectric:4.0")
 }
 
 addTestDependencies()
 
+val copyScript by tasks.registering(Copy::class) {
+    from(rootProject.file("../../packages/@nimbus-js/runtime/src/nimbus.js"))
+    into(file("src/main/assets/"))
```
threeve, line 38: my preference here is that rather than copying the file, we instead add the correct path to the asset srcDirs of the nimbus module.

paultiarks: okay, will change

paultiarks: Should be fixed now

PR #276, opened 2020-08-31
Test cases for parameter decoding error.

sfdctaka:
Adding test cases for decoding error. When decoding fails, e.g. a string is passed when a number is expected, the promise in Javascript will be rejected and therefore can be handled in `catch(error => {...})`.

This PR does not unify `error` object passed to the handler in `catch` between web view and JSCore. Web view uses `DecodeError` which is our custom `Error` struct where as JSCore uses `DecodingError` that is out-of-the-box from Apple. Unification of error class is out of the scope for this work.

An edge case was discovered while working on this PR. Suppose there is a native function that is expecting a `String` parameter. Then from Javascript a simple simple Javascript object is passed instead, like `{aaa:`bbb`}`. Decoding of such parameter fails in JSCore, however, in web view, the script handler passes this Javascript object as JSON string. Combined with `try? JSONDecoder().decode(type, from: data)` where `data` is data representation of that JSON string the code was failing to detect wrong type being passed. You can see how I handle that below in `WebViewConnection.swift`. Similarly, `null` being passed was failing to be detected so I fixed that too.

### DIFF
--- a/packages/test-www/test/shared-tests.js
+++ b/packages/test-www/test/shared-tests.js
@@ -686,6 +686,9 @@ function verifyStringDecoderRejectsBool() {
   });
 }
 
+// TODO: This function is not tested from native iOS code yet. When a simple JSON notation object
+// {aaa:"bbb"} crosses the bridge it is first stringified. The stringified JSON object can't be
+// decoded by Swift to generic Dictionary.
 function verifyStringDecoderRejectsObject() {
   __nimbus.plugins.testPlugin.takesString({ aaa: "bbb" }).catch((error) => {
     __nimbus.plugins.expectPlugin.pass();
--- a/packages/test-www/test/shared-tests.js
+++ b/packages/test-www/test/shared-tests.js
@@ -669,3 +669,175 @@ function verifyPromiseResolvesWithEncodableException2() {
 }
 
 // endregion
+
+// region decoder
+
+function verifyStringDecoderRejectsInt() {
+  __nimbus.plugins.testPlugin.takesString(5).catch((error) => {
+    __nimbus.plugins.expectPlugin.pass();
+    __nimbus.plugins.expectPlugin.finished();
+  });
+}
+
+function verifyStringDecoderRejectsBool() {
+  __nimbus.plugins.testPlugin.takesString(true).catch((error) => {
+    __nimbus.plugins.expectPlugin.pass();
+    __nimbus.plugins.expectPlugin.finished();
+  });
+}
+
+function verifyStringDecoderRejectsObject() {
+  __nimbus.plugins.testPlugin.takesString({ aaa: "bbb" }).catch((error) => {
+    __nimbus.plugins.expectPlugin.pass();
+    __nimbus.plugins.expectPlugin.finished();
+  });
+}
+
+function verifyStringDecoderRejectsNull() {
+  __nimbus.plugins.testPlugin.takesString(null).catch((error) => {
+    __nimbus.plugins.expectPlugin.pass();
+    __nimbus.plugins.expectPlugin.finished();
+  });
+}
+
+function verifyStringDecoderRejectsUndefined() {
+  var und;
+  __nimbus.plugins.testPlugin.takesString(und).catch((error) => {
+    __nimbus.plugins.expectPlugin.pass();
+    __nimbus.plugins.expectPlugin.finished();
+  });
+}
+
+function verifyNumberDecoderRejectsString() {
+  __nimbus.plugins.testPlugin.takesNumber("number").catch((error) => {
+    __nimbus.plugins.expectPlugin.pass();
+    __nimbus.plugins.expectPlugin.finished();
+  });
+}
+
+function verifyNumberDecoderRejectsObject() {
+  __nimbus.plugins.testPlugin.takesNumber({ aaa: "bbb" }).catch((error) => {
+    __nimbus.plugins.expectPlugin.pass();
+    __nimbus.plugins.expectPlugin.finished();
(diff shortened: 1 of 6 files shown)


platforms/apple/Sources/Nimbus/WebViewConnection.swift
```diff
@@ -67,9 +67,26 @@ public class WebViewConnection: Connection, CallableBinder {
...
+            do {
+                let value = try JSONDecoder().decode(type, from: data)
+                return .success(value)
+            } catch {
+                if let decodingError = error as? DecodingError {
+                    // This is a special case we need to catch&handle where the input is a string that
+                    // contains JSON but not of a particular user defined data type.
+                    if case .typeMismatch(let incomingType, let context) = decodingError {
+                        if incomingType is String.Type &&
+                            context.debugDescription == "Expected to decode String but found a dictionary instead." {
```
sfdctaka, line 80: Detecting the type mismatch where JSON string was passed instead of a string as I found was by looking at the `incomingType` and `context.debugDescription`. I didn't want to use a string phrase for this detection but there was nothing else that I could use.

platforms/apple/Sources/Nimbus/WebViewConnection.swift
```diff
@@ -67,9 +67,39 @@ public class WebViewConnection: Connection, CallableBinder {
...
+                if let decodingError = error as? DecodingError {
+                    // This is a special case we need to catch&handle where the input is a string that
+                    // contains JSON but not of a particular user defined data type.
+                    if case let .typeMismatch(incomingType, context) = decodingError {
+                        if incomingType is String.Type,
+                            context.debugDescription == "Expected to decode String but found a dictionary instead." {
+                            return .failure(error)
+                        }
+                    }
+                    if #available(iOS 13, macOS 10.15, *) {
```
threeve, line 84: I know it's an extremely edgy edge case, however this code path now makes it impossible to pass the string literal `"null"` as the value to a string parameter, because `JSONDecoder` will treat it as a missing `String` value, however it would have succeeded had it fallen through to the `as? T` branch below.

sfdctaka:
@threeve I observed that removing `debugDescription` from the condition actually suffices to determine whether a JSON notation object in string is a legit string or not.

I also attempted various approaches to solve the problem of identifying whether `"null"` is null or a string but this one turns out to be very difficult. I'm not sure if this is doable. Because both nulls reach our `decode` method as quoted string there is no inherent marker that could be used to make the distinction whether it is a null or a string.

As you saw it the code in my PR will fail on `"null"`. I tried the code in `main` branch and the `decode` method there will fail on `null` instead.

So my question now is, should we support `null` at all? And just keep supporting `"null"` as we have always been doing?

platforms/apple/Sources/Nimbus/WebViewConnection.swift
```diff
@@ -67,9 +67,19 @@ public class WebViewConnection: Connection, CallableBinder {
...
-            return .success(value)
+        if let data = data {
+            do {
+                let value = try JSONDecoder().decode(type, from: data)
+                return .success(value)
+            } catch {
+                // This is a special case we need to catch&handle where the input is a string that
+                // contains JSON but not of a particular user defined data type.
+                if let decodingError = error as? DecodingError,
+                    case let .typeMismatch(incomingType, _) = decodingError,
```
threeve, line 78: Unfortunately this is not going to work still. This will make it impossible to pass a string as a string if the value of the string contains valid JSON.

E.g., if you try to pass the string value `"[]"` it will fail here and not fall through to the cast below.

threeve (changes requested):

platforms/apple/Sources/Nimbus/WebViewConnection.swift
sfdctaka: I will file a work item so we can work on RFC to resolve either by stringifying all parameters first or adding by new metadata approach.

sfdctaka:
Removed the test `testVerifyStringDecoderRejectsObject` and also the `do{try{...}}catch{...}` code block that attempted to handle the scenario where stringified JSON object failed to decode to generic `Dictionary`.

packages/test-www/test/shared-tests.js
```diff
@@ -669,3 +669,184 @@ function verifyPromiseResolvesWithEncodableException2() {
...
+}
+
+function verifyStringDecoderRejectsBool() {
+  __nimbus.plugins.testPlugin.takesString(true).catch((error) => {
+    __nimbus.plugins.expectPlugin.pass();
+    __nimbus.plugins.expectPlugin.finished();
+  });
+}
+
+function verifyStringDecoderRejectsObject() {
```
paultiarks, line 689: It looks like this test isn't being called?

paultiarks (changes requested):

packages/test-www/test/shared-tests.js
threeve: This is the test case that cannot currently pass on iOS. I think we should at least include a comment here to that effect.

paultiarks (approved):

platforms/apple/Sources/Nimbus/WebViewConnection.swift
sfdctaka: Per conversation, punting.

packages/test-www/test/shared-tests.js
sfdctaka: I added a comment. Tell me if you rather just remove the function.

threeve (approved):

PR #277, opened 2020-09-02
Enable native to call Javascript function without a return value.

sfdctaka:
* Added overloaded `evaluate` method for web view bridge and JSContext bridge.
* Web view refactored to fix overloaded `evaluate`.

### DIFF
--- a/platforms/apple/Sources/Nimbus/WebViewBridge.swift
+++ b/platforms/apple/Sources/Nimbus/WebViewBridge.swift
@@ -26,7 +26,7 @@ public class WebViewBridge: NSObject, JSEvaluating {
      Invokes a Promise-returning Javascript function and call the specified
      promiseCompletion when that Promise resolves or rejects.
      */
-    func invoke<R>(
+    func invoke<R>( // swiftlint:disable:this function_body_length
         _ identifierSegments: [String],
         with args: [Encodable],
         callback: @escaping (Error?, R?) -> Void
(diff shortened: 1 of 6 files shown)


platforms/apple/Sources/Nimbus/WebViewBridge.swift
```diff
@@ -56,15 +56,51 @@ public class WebViewBridge: NSObject, JSEvaluating {
...
             return callback(error, nil)
         }
 
+        evaluateJavaScript(promiseId: promiseId, idSegmentString: idSegmentString, argString: argString)
+    }
+
+    /**
+     Overloaded version of invoke that is used to call a Javascript function that doesn't have a return value.
+     */
+    func invoke(
```
threeve, line 70: Why is web view bridge using a different approach than the js bridge? It seems like you could do the same thing to wrap the callback and drop the result and then no longer need this additional `invoke` overload?

platforms/apple/Sources/NimbusTests/CallJavascriptTests.swift
```diff
@@ -162,6 +178,17 @@ class CallJSContextTests: XCTestCase {
         XCTAssertEqual(resultValue, true)
     }
 
+    func testCallFunctionWithoutReturn() throws {
+        let expect = expectation(description: "test call function without return")
+        var error: Error?
+        bridge!.invoke(["testFunctionWithoutReturn"]) { theError, _ in
```
threeve, line 184: Should this be testing the no-return version of `evaluate` instead of `invoke` here?

threeve (approved):

PR #279, opened 2020-09-03
W-8040180: Nimbus generated v8 binders need to execute on the v8 executor

gatorboy:
https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008XdfuIAC/view

Added executor to the V8Bridge
Updated tests.
Pending fix for PrimitiveV8EncodableTests.testBooleanToV8

### DIFF
--- a/platforms/android/demo-app/src/main/java/com/salesforce/nimbusdemoapp/MainActivity.kt
+++ b/platforms/android/demo-app/src/main/java/com/salesforce/nimbusdemoapp/MainActivity.kt
@@ -7,6 +7,7 @@
 
 package com.salesforce.nimbusdemoapp
 
+import android.app.Activity
 import android.content.Context
 import android.os.Bundle
 import android.util.Log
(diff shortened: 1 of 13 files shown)


jamie-houston:
This'll be another breaking change (executor in constructor for V8Bridge) so we'll need to be careful about pushing this update as it'll break all the consumers again.

threeve:
@jamie-houston the only people using v8 bridge right now is us, as far as I know. Would it be better if we default the argument to a new `singleThreadExecutor`?

jamie-houston:
@threeve good point about no other consumers.  In that case I think it's fine to make this change now.  Regarding a default `singleThreadExecutor` I'm trying to think of the consequence of that and if it'll cause hard to track down bugs.  Maybe leave it required for now.  Changing to optional later would be backwards compatible.

jamie-houston (approved):

PR #282, opened 2020-09-10
Android parameter rejection tests

paultiarks:
This adds the same tests for Android that were recently added for iOS to test invoking bound functions with parameters that should cause them to be rejected.

A number of scenarios were discovered that are currently not possible to support on Android without further work. These scenarios are currently commented out and I will create a story for the backlog to come up with a solution for both Android and iOS that can be implemented in a future version.

### DIFF
--- a/platforms/android/shared-tests/src/androidTest/java/com/salesforce/nimbus/bridge/tests/v8/V8PluginTests.kt
+++ b/platforms/android/shared-tests/src/androidTest/java/com/salesforce/nimbus/bridge/tests/v8/V8PluginTests.kt
@@ -381,9 +381,128 @@ class V8PluginTests {
 
     // endregion
 
-    private fun executeTest(function: String) {
+    // region parameter errors
+
+    @Test
+    fun testVerifyStringDecoderRejectsInt() {
+        executeTest("verifyStringDecoderRejectsInt()")
+    }
+
+    @Test
+    fun testVerifyStringDecoderRejectsBool() {
+        executeTest("verifyStringDecoderRejectsBool()")
+    }
+
+    @Test
+    fun testVerifyStringDecoderRejectsNull() {
+        executeTest("verifyStringDecoderRejectsNull()")
+    }
+
+    @Test
+    fun testVerifyStringDecoderRejectsUndefined() {
+        executeTest("verifyStringDecoderRejectsUndefined()")
+    }
+
+    @Test
+    fun testVerifyStringDecoderResolvesStringNull() {
+        executeTest("verifyStringDecoderResolvesStringNull()")
+    }
+
+    @Test
+    fun testVerifyNumberDecoderRejectsString() {
+        executeTest("verifyNumberDecoderRejectsString()")
+    }
+
+    @Test
+    fun testVerifyNumberDecoderRejectsObject() {
+        executeTest("verifyNumberDecoderRejectsObject()")
+    }
+
+    @Test
+    fun testVerifyNumberDecoderRejectsNull() {
+        executeTest("verifyNumberDecoderRejectsNull()")
+    }
+
+    @Test
+    fun testVerifyNumberDecoderRejectsUndefined() {
+        executeTest("verifyNumberDecoderRejectsUndefined()")
+    }
+
+    @Test
+    fun testVerifyBoolDecoderRejectsString() {
+        executeTest("verifyBoolDecoderRejectsString()")
+    }
+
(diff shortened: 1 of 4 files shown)


platforms/android/shared-tests/src/androidTest/java/com/salesforce/nimbus/bridge/tests/webview/WebViewPluginTests.kt
```diff
@@ -443,10 +443,140 @@ class WebViewPluginTests {
...
+        executeTest("verifyTestStructDecoderRejectsUndefined()")
+    }
+
+    // endregion
+
     private fun executeTest(script: String) {
         expectPlugin.testReady.withTimeoutInSeconds(30) {
             runOnUiThread { webView.evaluateJavascript(script) {} }
-            assertThat(expectPlugin.testFinished.await(30, TimeUnit.SECONDS)).isTrue()
+            assertThat(expectPlugin.testFinished.await(3, TimeUnit.SECONDS)).isTrue()
```
threeve, line 579: intentional? or was this supposed to just be a local change to speed up dev?

paultiarks: A little of both? 30 is definitely way too high, even for firebase/CI.

paultiarks: I'm putting the 30 back though, just to not mess with it.

platforms/android/shared-tests/src/androidTest/java/com/salesforce/nimbus/bridge/tests/v8/V8PluginTests.kt
```diff
@@ -381,9 +381,128 @@ class V8PluginTests {
...
+    }
+
+    @Test
+    fun testVerifyTestStructDecoderRejectsNull() {
+        executeTest("verifyTestStructDecoderRejectsNull()")
+    }
+
+    @Test
+    fun testVerifyTestStructDecoderRejectsUndefined() {
+        executeTest("verifyTestStructDecoderRejectsUndefined()")
```
jamie-houston, line 498: one day we'll change these tests to use reflection or something a little easier to manage.

platforms/android/shared-tests/src/androidTest/java/com/salesforce/nimbus/bridge/tests/webview/WebViewPluginTests.kt
```diff
@@ -443,6 +443,136 @@ class WebViewPluginTests {
 
     // endregion
 
+    // region parameter errors
+
+    // TODO: Currently not possible on Android
+//    @Test
```
jamie-houston, line 449: Change to `@Ignore("Currently not possible on Android")`

Is there a work item we can add to them as well?

paultiarks: ahh good call, will create one and add it

platforms/android/shared-tests/src/androidTest/java/com/salesforce/nimbus/bridge/tests/v8/V8PluginTests.kt
```diff
@@ -381,9 +381,128 @@ class V8PluginTests {
...
+    }
+
+    // endregion
+
+    private fun executeTest(funtion: String) {
         assertThat(expectPlugin.testReady.await(30, TimeUnit.SECONDS)).isTrue()
-        bridge.executorScope(executorService) { bridge.executeScriptOnExecutor(function) }
+        bridge.executorScope(executorService) { bridge.executeScriptOnExecutor(funtion) }
         assertThat(expectPlugin.testFinished.await(30, TimeUnit.SECONDS)).isTrue()
```
jamie-houston, line 506: Was this on purpose...?  Can we change to a properly spelled word?  Not sure if `function` is taken, but we could use `testFunction` or something if it is...?

paultiarks: this was definitely not on purpose.

jamie-houston (approved):

PR #285, opened 2020-09-16
setTimeout for JSCore

sfdctaka:
* Touch up on existing `setTimeout` and `clearTimeout`. Moved it into nimbus core from test target.
* Added `setInterval` and `clearInterval`.
* Added shared tests to be shared with Android.


packages/test-www/test/shared-tests.js
```diff
@@ -853,3 +853,53 @@ function verifyTestStructDecoderRejectsUndefined() {
...
+  });
+  testPlugin.wait(5000).then(()=>{
+    if (called == 3) {
+      __nimbus.plugins.expectPlugin.pass();
+      __nimbus.plugins.expectPlugin.finished();
+    }
+  });
+}
+
+// endregion
```
sfdctaka, line 905: @nadeem-sfdc Here are tests that can be reused with Android.

platforms/apple/Sources/Nimbus/BridgeBuilder.swift
```diff
@@ -19,6 +19,7 @@ public class BridgeBuilder {
 
     public static func createBridge(for context: JSContext, plugins: [Plugin]) -> JSContextBridge {
         let bridge = JSContextBridge(context: context, plugins: plugins)
+        JSContextGlobals.setupGlobals(context: context)
```
sfdctaka, line 22: Setting up `setTimeout` and related functions are now done inside creation of bridge so that these functions come together out-of-the-box with jscontext.

platforms/apple/Sources/NimbusTests/SharedTestsJSCore.swift
```diff
@@ -41,7 +42,7 @@ class SharedTestsJSCore: XCTestCase {
     func executeTest(_ testName: String) {
         XCTAssertTrue(expectPlugin.isReady)
         context.evaluateScript(testName)
-        waitForExpectations(timeout: 1, handler: nil)
+        waitForExpectations(timeout: 10, handler: nil)
```
sfdctaka, line 45: Bumped up timeout so that test for interval can easily fit in.

PR #287, opened 2020-10-01
Make JSValue encoder and decoder public.

sfdctaka:
These methods can be extremely beneficial to LMR Context that is a wrapper object containing JSCore.


paultiarks (commented):
I'm a little hesitant to make this part of the Nimbus public API since part of the goal of Nimbus is to handle the serialization/deserialization across the bridge and never have to deal with `JSValue`s directly. I wonder if there is a better solution to the problem attempted to be solved here. I'll defer to @threeve

threeve:
@paultiarks I think perhaps for a future version of Nimbus we're gonna want to add a way for Nimbus to support more use cases, but for now I'm ok making these public so they can be used outside of the Nimbus bridge itself.

threeve (approved):

PR #294, opened 2020-10-06
Fix circle jobs that had identical names

paultiarks:
I think having these two tasks named the same was causing only the last one in the file to get parsed and therefore only working on pre-releases. This should fix that.


jamie-houston (approved):

threeve (approved):
🤞

PR #299, opened 2020-10-22
reverting package name with bintray fix

jamie-houston:
Bintray has updated the original bintray package (https://bintray.com/salesforce-mobile/android/nimbus) to publish to jcenter, so this change to package name is not longer necessary.  Future publications to the package 'nimbus' should also publish to jcenter.


threeve (approved):
lgtm

PR #300, opened 2020-11-05
@W-8291379:  primarily fix the crash issue when bound method return type is Unit

ben-zhang-at-salesforce:
did two fixes
1. fix the crash issue when bound method return type is Unit
2. fix reference leak issue for promise resolve/reject

### DIFF
--- a/platforms/android/modules/bridge-v8/src/androidTest/java/com/salesforce/nimbus/bridge/v8/V8ExtensionsTest.kt
+++ b/platforms/android/modules/bridge-v8/src/androidTest/java/com/salesforce/nimbus/bridge/v8/V8ExtensionsTest.kt
@@ -1,0 +1,116 @@
+//
+// Copyright (c) 2020, Salesforce.com, inc.
+// All rights reserved.
+// SPDX-License-Identifier: BSD-3-Clause
+// For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+//
+
+package com.salesforce.nimbus.bridge.v8
+
+import com.eclipsesource.v8.V8
+import io.mockk.every
+import io.mockk.mockk
+import io.mockk.slot
+import org.junit.After
+import org.junit.Assert
+import org.junit.Before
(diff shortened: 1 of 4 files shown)


platforms/android/modules/bridge-v8/src/androidTest/java/com/salesforce/nimbus/bridge/v8/V8ExtensionsTest.kt
```diff
@@ -0,0 +1,115 @@
...
+ * Unit tests for [V8Extensions].
+ */
+class V8ExtensionsTest {
+
+    lateinit var mV8: V8
+
+    /**
+     * mocked v8 executor service to make v8 to run on test thread to avoid test flaking due to timing
+     */
+    val mockedV8ExecutorService = mockk<ScheduledExecutorService> {
```
jamie-houston, line 31: Should we just make mockedV8ExecutorService in j2v8 public and re-use TestBase in https://git.soma.salesforce.com/MobilePlatform/j2v8-inspector/blob/main/j2v8-inspector/src/androidTest/java/com/salesforce/extension/TestBase.kt#L31?

jamie-houston (approved):
one question, but lgtm

platforms/android/modules/bridge-v8/src/androidTest/java/com/salesforce/nimbus/bridge/v8/V8ExtensionsTest.kt
ben-zhang-at-salesforce: let's have a discussion about the dependency, and which way to go.  like mockedV8ExecutorService and the V8.scope in j2v8, the V8.memoryScope I added  here and the V8.extension from j2v8 inspect like console, timeout. ... are all kind of common extension for j2v8. maybe we should structure like below ?

j2v8 
   --> j2v8 extension (basic function improvement and tests)
          ---> j2v8 Inspector
          ---> Nimbus 
                 .....

let's have a discussion monday

PR #304, opened 2020-11-13
reduce reference leak issue for nimbus v8 binders - to be divided into small piece for easy review

ben-zhang-at-salesforce:
reduce reference leak issue for nimbus and nimbus generated binder code. for sample lmr app, 142 references are properly removed.


threeve:
@ben-zhang-at-salesforce can you post a diff of the generated binder code before and after this change? It'd make it easier to review the changes to the binder generator code. thanks.

ben-zhang-at-salesforce:
splitted into two small PRs for easy code review.

PR #311, opened 2021-02-01
Update js and ruby dependencies

threeve:


threeve:
@nadeem-sfdc quick, give me a 👍 before the tests go wonky again 🤣

nadeem-sfdc:
Looks good to me. I didn't know if we had to update the -lock.json files too.

nadeem-sfdc (approved):

